Issue #2, opened 2023-04-03
Modify SuperGenius Finality with Hedera HashGraph's median Timestamp.


SuperGNUS:
Currently our consensus finality of a block is using a multi-round voting system.  Integrate HashGraphs median timestamp (probably requires some block structure changes), which will allow for a single-round voting system.

There may be some work to make sure the gossip protocol works with the new voting mechanism.

creativeid00:
https://docs.hedera.com/hedera/core-concepts/hashgraph-consensus-algorithms

creativeid00:
Hedera consensus node design https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/docs/design/design.md#architecture

creativeid00:
Hashgraph consensus algorithm whitepaper https://www.swirlds.com/downloads/SWIRLDS-TR-2016-01.pdf

Super-Genius:
Closed by inbox triage: stale issue from SuperGNUS era (1000+ days). No recent activity or updates. Please reopen with updated requirements if still relevant.

PR #24, opened 2023-06-09
Dev vote ordering

luizcarvalhohen:
Added timestamp in Vote messages - this can be used for ordering in median calculation in consensus algorithm.


Super-Genius (changes requested):
We need tests to be part of this commit before we merge.

Super-Genius:
Also, please put "closes #2" in the merge request so that it will close the task associated with the merge

PR #26, opened 2023-07-19
Dev fix tests

henriqueaklein:
- Changed one byte of the block hash so that the vote was not evaluated as a duplicate.
- Made the header encoding test pass. It didn't because there was no case and the GetHeader didn't ran.
- Reviewed Coding standard and set clang format accordingly (let me know if there's something non compliant to the coding standard.


Super-Genius (approved):

Super-Genius:
Closes #26

PR #54, opened 2024-07-22
Dev warningcleanup

itsafuu:
Some warnings remain, specifically 
/Users/fuu/gnus/SuperGenius/src/account/TransactionManager.hpp:67:58: warning: private field 'last_trans_on_block_id'
Left this alone due to assumption we may intend to use this.

ld: warning: ignoring duplicate libraries 
Many of these occur. 

clang: warning: argument unused during compilation: '-Xlink=-force:multiple'
Might be able to suppress this or manage it in a way it's only used in libraries that need it.

/src/crypto/bip39/mnemonic.cpp:23:33: warning: 'codecvt_utf8_utf16<wchar_t>' is deprecated [-Wdeprecated-declarations]
/src/crypto/bip39/mnemonic.cpp:23:12: warning: 'wstring_convert<std::codecvt_utf8_utf16<wchar_t>>' is deprecated
I don't know if we even plan to use bip39

stb_image_write.h has some warnings, but we won't even be using it. 

src/base/hexutil.hpp:84:14: warning: shift count >= width of type
I feel like the shift beyond width of type is the entire point of the function, in order to detect that. 

src/storage/database_error.cpp:31:5: warning: all paths through this function will call itself
This should possibly be dealt, it seems like a test setup though.

### DIFF
--- a/build/CompilationFlags.cmake
+++ b/build/CompilationFlags.cmake
@@ -20,7 +20,9 @@ if("${CMAKE_CXX_COMPILER_ID}" MATCHES "^(AppleClang|Clang|GNU)$")
     add_flag(-Wno-unused-function) # prints too many useless warnings
     add_flag(-Wno-format-nonliteral) # prints way too many warnings from spdlog
     add_flag(-Wno-gnu-zero-variadic-macro-arguments) # https://stackoverflow.com/questions/21266380/is-the-gnu-zero-variadic-macro-arguments-safe-to-ignore
-
+	add_flag(-Wno-unused-result) #Every logger call generates this
+	add_flag(-Wno-pessimizing-move) #Warning was irrelevant to situation
+	
     # promote to errors
     # add_flag(-Werror=unused-lambda-capture)  # error if lambda capture is unused
     # add_flag(-Werror=return-type)      # warning: control reaches end of non-void function [-Wreturn-type]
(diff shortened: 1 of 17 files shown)


build/CompilationFlags.cmake
```diff
@@ -20,7 +20,9 @@ if("${CMAKE_CXX_COMPILER_ID}" MATCHES "^(AppleClang|Clang|GNU)$")
     add_flag(-Wno-unused-function) # prints too many useless warnings
     add_flag(-Wno-format-nonliteral) # prints way too many warnings from spdlog
     add_flag(-Wno-gnu-zero-variadic-macro-arguments) # https://stackoverflow.com/questions/21266380/is-the-gnu-zero-variadic-macro-arguments-safe-to-ignore
-
+	add_flag(-Wno-unused-result) #Every logger call generates this
```
henriqueaklein, line 23: Disabling unused result might hide some bad practices. Do we have a way to change the logger to avoid this? If not, it's fine then.

build/CompilationFlags.cmake
```diff
@@ -20,7 +20,9 @@ if("${CMAKE_CXX_COMPILER_ID}" MATCHES "^(AppleClang|Clang|GNU)$")
...
+	add_flag(-Wno-pessimizing-move) #Warning was irrelevant to situation
```
henriqueaklein, line 24: This disable std::move pessimizing. Does our code has that or thirdparty non-forked repos? If it's ours we could evaluate where it happens instead

henriqueaklein (approved):

PR #58, opened 2024-09-23
GeniusWallet integration

EduMenges:
To be merged with https://github.com/GeniusVentures/GeniusWallet/pull/77 and https://github.com/GeniusVentures/GeniusSDK/pull/9

### DIFF
--- a/build/CommonBuildParameters.cmake
+++ b/build/CommonBuildParameters.cmake
@@ -80,13 +80,10 @@ set(OPENSSL_LIBRARIES "${OPENSSL_DIR}/lib" CACHE PATH "Path to OpenSSL lib folde
 set(OPENSSL_CRYPTO_LIBRARY ${OPENSSL_LIBRARIES}/libcrypto${CMAKE_STATIC_LIBRARY_SUFFIX} CACHE PATH "Path to OpenSSL crypto lib")
 set(OPENSSL_SSL_LIBRARY ${OPENSSL_LIBRARIES}/libssl${CMAKE_STATIC_LIBRARY_SUFFIX} CACHE PATH "Path to OpenSSL ssl lib")
 find_package(OpenSSL REQUIRED)
-include_directories(${OPENSSL_INCLUDE_DIR})
 
 # rocksdb
 set(RocksDB_DIR "${_THIRDPARTY_BUILD_DIR}/rocksdb/lib/cmake/rocksdb")
-set(RocksDB_INCLUDE_DIR "${_THIRDPARTY_BUILD_DIR}/rocksdb/include")
 find_package(RocksDB CONFIG REQUIRED)
-include_directories(${RocksDB_INCLUDE_DIR})
 
 # stb
 include_directories(${_THIRDPARTY_BUILD_DIR}/stb/include)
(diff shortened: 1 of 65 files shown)


src/account/GeniusNode.hpp
```diff
@@ -78,6 +78,21 @@ namespace sgns
         void     AddPeer( const std::string &peer );
         uint64_t GetBalance();
 
+        [[nodiscard]] const std::vector<std::vector<uint8_t>> &GetTransactions() const
+        {
+            return transaction_manager_->GetTransactions();
+        }
+
+        std::string GetAdddress()
```
matthewhoman, line 86: 3 d's here on address

src/account/UTXOTxParameters.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    int64_t remain = total_amount;
+
+    for ( auto &utxo : utxo_pool )
+    {
+        if ( remain <= 0 )
+        {
+            break;
+        }
+        InputUTXOInfo curr_input{ utxo.GetTxID(), utxo.GetOutputIdx(), std::move( signature ) };
```
henriqueaklein, line 24: If you std::move this it will lose the value the first iteration.

henriqueaklein (changes requested):

src/account/UTXOTxParameters.cpp
EduMenges: 🤡

build/CommonBuildParameters.cmake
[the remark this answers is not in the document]
EduMenges: The target that includes the correct things in `AsyncIOManagerConfig.cmake` is actually `ipfs-unixfs`. I did not bother fixing it, so I kept the `INCLUDE_DIR`.

henriqueaklein (approved):

itsafuu:
I believe it would be better to return transactions from GeniusSDKGetTransactions in a different way. 
Propose:
First, de-serialize the transaction data with protobuf before returning it. Currently the dashboard/wallet need protobuf to handle reading the data that comes back. This can be done with SGTransactions.pb.h 

Second, include the key for each transaction for more easily detecting dupes when refreshing transactions. 


Additionally, another method in here for getting blockchain header data/blockchain payload data. Again, de-serialize with SGBlocks.pb.h.

EduMenges:
> First, de-serialize the transaction data with protobuf before returning it. Currently the dashboard/wallet need protobuf to handle reading the data that comes back. This can be done with SGTransactions.pb.h

But then how would the wallet read the data? We initially passed the raw protobuf bytes to it because the transactions are a tagged union and the whole desserialization thing is better left up to protobuf.

itsafuu:
> > First, de-serialize the transaction data with protobuf before returning it. Currently the dashboard/wallet need protobuf to handle reading the data that comes back. This can be done with SGTransactions.pb.h
> 
> But then how would the wallet read the data? We initially passed the raw protobuf bytes to it because the transactions are a tagged union and the whole desserialization thing is better left up to protobuf. I could create a custom C struct and use that, but I think we would just be re-implementing the protobuf part.

Yes, a custom C struct is pretty much in line with what I was thinking. We're currently in a situation where any app linking to the SDK that wants to read transactions must also have protobuf files to decode the information. Not to mention because the transaction could be a mint, transfer, escrow, we must do a bunch of try catch to check which type of de-serialization we need. Though providing the "key" would allow us to check that before we try to de-serialize. 

Either way, I don't know how many use cases of reading transaction data apps linking to the SDK we might have, but expecting anyone who wants to do that to grab .proto files and convert them into an appropriate protobuf reader for the language of their app seems to include a lot of friction. 

We may way to address that later though.


 The dashboard still needs those block headers/tx from the SDK for display. So if we're not going to de-serialize right now, you could return those anyway with a function to get blocks, I already have protobuf files in dashboard.

henriqueaklein:
I think as far as friction goes, either the person has to know how to walk around a custom C struct or deserialize proto in their own chosen language. Anyway, the access to transactions and stuff I think will be mostly by our stuff. A game dev will worry only about calling Init, so this change would be mostly for us.

I think we could have the custom C struct in GeniusSDK if it makes it any easier on dart. If we want to provide these APIs in a friction-less manner in other languages we could do what TrustWallet does for Java/Swift, that is to implement higher level APIs that does the desserialization for them.

EduMenges:
> We're currently in a situation where any app linking to the SDK that wants to read transactions must also have protobuf files to decode the information.

Yeah, that is a good point, but we also have to consider that not every language has a good C interop (like JavaScript), so maybe working with the C struct would make it harder for these cases. JS is dumb tho, probably the users are going to use SGNS with C#, C++ and other game engine friendly languages.

Another concern is dynamic collections: if we need to use vectors and strings, protobuf handles that part, we just need to pass the raw bytes and de-alloc them later. But if we are going to use a C struct, we would need to either make these collections have a static size or deal with allocation and size definition, which would increase complexity. The user would have to replicate this in their own language so that calling out function actually works.

itsafuu:
In the interest of getting things done quickly, for now I would like to see a method to get the blocks, and we can keep returning protobuf. It would also still be nice to also get the key value for the crdt entry along with these so I can avoid a bunch of try catch blocks for decoding with protobuf.

EduMenges:
> for now I would like to see a method to get the blocks

Do you mean to get a whole block from the blockchain?

itsafuu:
> > for now I would like to see a method to get the blocks
> 
> Do you mean to get a whole block from the blockchain?

Yeah, see screenshot I posted a couple days ago to general of the Dashboard, it has "Recent Blocks" and "Recent Tranasactions".

itsafuu:
> > Yeah, see screenshot I posted a couple days ago to general of the Dashboard, it has "Recent Blocks" and "Recent Tranasactions".
> 
> Ok, should we get the block and the header?

Yeah, I think in theory it displays the header and you could click on it to get the block info itself. Probably best to have both.

PR #81, opened 2025-02-04
dev_fixed_point to develop: fixed point operations on account

luizgrz:
## Added
- Implemented fixed-point arithmetic operations
- Added new unit tests to validate fixed-point operations

## Changed
- Refactored existing code to integrate fixed-point calculations
- Updated documentation to reflect changes in arithmetic operations

## Fixed
- Inconsistencies in calculations due to previous floating-point arithmetic

## Removed
- `extern "C"` since SDK modules are not compatible with pure C


src/account/GeniusNode.cpp
```diff
@@ -437,19 +447,37 @@ namespace sgns
...
+        }
+
+        return std::max( cost, static_cast<uint64_t>( 1 ) );
     }
 
-    outcome::result<void> GeniusNode::MintTokens( double      amount,
-                                                  std::string transaction_hash,
-                                                  std::string chainid,
-                                                  std::string tokenid )
+    outcome::result<void> GeniusNode::MintTokens( const std::string &amount_str,
```
henriqueaklein, line 463: I don't think its necessary for the mint to have a string. If its a uint64_t we expect it to be already in Minion Tokens, so no conversion needed.

test/src/multiaccount/multi_account_sync.cpp
```diff
@@ -112,16 +112,16 @@ sgns::GeniusNode *MultiAccountTest::node_main  = nullptr;
...
+DevConfig_st MultiAccountTest::DEV_CONFIG2 = { "0xcafe", "0.65", 1.0, 1, "./node2" };
+DevConfig_st MultiAccountTest::DEV_CONFIG3 = { "0xcafe", "0.65", 1.0, 0, "./node3" };
 
 std::string MultiAccountTest::binary_path = "";
 
 
 TEST_F( MultiAccountTest, SyncThroughThirdNode )
 {
-    node_main->MintTokens( 50, "", "", "" );
+    node_main->MintTokens( "50", "", "", "" );
```
henriqueaklein, line 124: From now on this will be represented in Minions, so it would be 50000000000 I think

henriqueaklein (changes requested):
This is all very good. Just set the interface of mint token to Minions and it should be fine

src/account/GeniusNode.cpp
luizgrz: Fixed on #[254ebf1](https://github.com/GeniusVentures/SuperGenius/pull/81/commits/254ebf19f28bba2f96476a4ae8c399a9c6404312)

test/src/multiaccount/multi_account_sync.cpp
luizgrz: Fixed on #[254ebf1](https://github.com/GeniusVentures/SuperGenius/pull/81/commits/254ebf19f28bba2f96476a4ae8c399a9c6404312)

PR #92, opened 2025-02-26
TX Check with Wait

Super-Genius:
Transactions now have check for processed transactions with timeout instead of sleep.  

Only transaction_sync_test was updated to remove sleeps, but it is getting a consistent GraphSync can't find id:1 error, so this is good to be able to hunt that bug down.


henriqueaklein:
It's not working because of the path of the notification, so WaitForTransactionIncoming never gets anything. I'm on it

henriqueaklein (approved):

PR #94, opened 2025-02-27
Get prices from coingecko for use in processing cost calculations

itsafuu:


henriqueaklein (approved):
Seems OK. I'll check it with asan just to make sure

PR #97, opened 2025-03-03
Fix weak ptr crdt issue

Super-Genius:
Fixing weak_ptr reference issue that would hold a strong refernce until the object references the lock was out of scope
  
We may need to look at all weak_ptr implementations for threads that are both creating a long term weak reference and a strong reference for threads.

This negates the weak_ptr intent.  See https://grok.com/share/bGVnYWN5_93190cc3-96d6-43ef-ac0d-ef4116fe6bec

Also crdt_tests are now passing.


henriqueaklein (changes requested):
transaction_sync and processing_nodes don't work anymore. Basically the changes made it that no request on graphsync works

PR #101, opened 2025-03-07
Fixes the queue timer being off between nodes.

Super-Genius:
This should thereby prevent double spending and loss of ability to process transactions,

The code hasn't been tested or built yet, but checking in for Eduardo or Henrique to do in the morning.


henriqueaklein:
Just tested it on the processing nodes, somehow it's worse, now both tasks grab the exact same subtask at the same time

henriqueaklein:
Implemented a timed request to be the one who creates the queue and it seemed to fix that, but double payouts happen still because one node has the queue and finishes the job and proceeds to payout. The second node requested the queue, the CRDT hasn't synched yet and it begins to payout

PR #106, opened 2025-03-13
Some memory optimization stuff

itsafuu:
Not really finished with this yet, just making PR for comparison


henriqueaklein (approved):

src/crdt/impl/crdt_datastore.cpp
```diff
@@ -824,7 +811,7 @@ namespace sgns::crdt
                                         aRoot.toString().value() );
                         return outcome::failure( replaceResult.error() );
                     }
-
+                    AddProcessedCID(child);
```
henriqueaklein, line 814: This seems to be correct but as we discussed briefly, we could stop using this since the merkledag syncher seems to be working now, which is basically where we store any CIDs we successfully grab from graphsync on the Block received callback on graphsyncdagsyncher

test/src/pubsub_counts/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+
+addtest(pubsub_counts_test
```
henriqueaklein, line 2: The CI seems to be erroring on gtest so I guess this new test must be the reason why

henriqueaklein (changes requested):
Fix the gtest errors

PR #116, opened 2025-04-10
Use the pubsub libp2p host for graphsyncs as well to simplify networking

itsafuu:
Currently fear what the result will be when multi globaldb uses a single graphsync, will they collide with each other? 

For msvc in globaldb_integration_gtest used #define WAIT_TIMEOUT ( std::chrono::milliseconds( 10000 ) ) because it was unwilling to accept any alternative.


luizgrz:
I’m going to resolve some conflicts and start testing locally

PR #122, opened 2025-05-02
Dev crdt filter + multi topics + graphsync integration

henriqueaklein:
This is a huge one. Basically 3 different works were merged into one.
- CRDT Filter:
  - Created filter for CRDT incoming deltas. If the content is rejected, a tombstone can be created for that and other deltas.
  - Simplified transaction keys. Removed the need for replicates and created better querying on CRDT/rocksdb
  - Improvements on speed and memory usage
  - Fixed CRDT sync when boot up not to grab every block.
- Multi  topics:
  - Multiple topics for broadcasting/listening on the broadcaster.
  - Possible to select topics only for listening and broadcasting. Unified globaldb into one. 
  - Reduced number of threads from CRDT from 7 for each, summing up to 15 at least to 3 total.
  - Improved handling of crdt threads using condition variable
  - Removed duplicated messages received by broadcaster.
- Graphsyn integration:
  - Removed need for graphsync port, using the same as pubsub.
  - Dealt with many issues on cleanup (resource de-allocation)
  - Enables creating multiple CRDTs (not done yet) with same graphsync


src/account/GeniusNode.cpp
```diff
@@ -148,7 +147,7 @@ namespace sgns
         loggerGraphsync->set_level( spdlog::level::err );
         loggerBroadcaster->set_level( spdlog::level::err );
         loggerDataStore->set_level( spdlog::level::err );
-        loggerTransactions->set_level( spdlog::level::err );
+        loggerTransactions->set_level( spdlog::level::debug );
```
luizgrz, line 150: maybe revert the default to error on develop

src/account/GeniusNode.cpp
```diff
@@ -185,30 +184,62 @@ namespace sgns
...
             {
-                node_logger->error( "Failed to open port" );
+                int candidate_port = pubsubport + i;
+                if ( upnp->CheckIfPortInUse( candidate_port, "TCP", owner ) )
+                {
+                    if ( owner == lanip )
+                    {
+                        node_logger->info( "Port {} is already mapped by this device. Using it.", candidate_port );
+                        addresses.push_back( wanip );
+                        success = true;
```
luizgrz, line 213: pubsubport is not updated with candidate_port?

src/account/GeniusNode.cpp
```diff
@@ -185,30 +184,62 @@ namespace sgns
...
+                {
+                    node_logger->warn( "Failed to open port {}", candidate_port );
+                }
             }
-            else
+
+            if ( !success )
             {
-                node_logger->info( "Open Port Success" );
+                node_logger->error( "Unable to open a usable UPnP port after {} attempts", max_attempts );
```
luizgrz, line 239: We can move forward here?

src/account/GeniusNode.cpp
```diff
@@ -225,34 +256,41 @@ namespace sgns
...
+                                                      scheduler,
+                                                      generator );
+        globaldb_->AddBroadcastTopic( processing_channel_topic_ );
+        globaldb_->AddListenTopic( processing_channel_topic_ );
         if ( global_db_init_result.has_error() )
         {
             auto error = global_db_init_result.error();
             throw std::runtime_error( error.message() );
             return;
         }
```
luizgrz, line 292:
```suggestion
        if ( global_db_init_result.has_error() )
        {
            auto error = global_db_init_result.error();
            throw std::runtime_error( error.message() );
            return;
        }
        globaldb_->AddBroadcastTopic( processing_channel_topic_ );
        globaldb_->AddListenTopic( processing_channel_topic_ );
```

luizgrz:
Can we remove AddProcessedCID ,  ContainsCID and DeleteCIDS?

src/account/GeniusNode.cpp
itsafuu: Yes, even if we fail to open a port chances are outgoing connections can still happen, though it may make the node less effective at syncing. Future libp2p updates will also include circuit relay and holepunching.

PR #125, opened 2025-05-16
Dev fix processing topic

henriqueaklein:
- Created second CRDT instance with the same rocksdb storage in order to separate broadcasts of transactions that were being sent to processing topics.
- Delayed initialization/Start of crdt datastore and globaldb to enable filter registering before starting everything.
- Cleanup of useless checks and some dead code.
- Changed crdt filter to class instead of static methods.
- Simplified globaldb creation.


src/crdt/impl/crdt_data_filter.cpp
```diff
@@ -0,0 +1,85 @@
...
+        for ( const auto &tombstone : new_tombstones )
+        {
+            auto *new_tombstone = delta->add_tombstones();
+            *new_tombstone      = tombstone;
+        }
+    }
+
+    void CRDTDataFilter::FilterTombstonesOnDelta( std::shared_ptr<pb::Delta> &delta )
+    {
+        //TODO - Figure out how to remove tombstones even recorded ones
```
luizgrz, line 83: Maybe throw an exception "not supported"

src/crdt/impl/crdt_data_filter.cpp
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        for ( int i = 0; i < delta->elements_size(); ++i )
+        {
+            const auto &element        = delta->elements( i );
+            bool        filter_matched = false;
+
+            for ( const auto &[pattern, filter] : registry_copy )
+            {
+                std::regex regex( pattern );
```
luizgrz, line 55: The regex for this filter could not be generated before and stored with the ElementFilterCallback?

src/crdt/impl/crdt_data_filter.cpp
```diff
@@ -0,0 +1,85 @@
...
+        std::lock_guard lock( tombstone_registry_mutex_ );
+        tombstone_registry_.erase( pattern );
+    }
+
+    void CRDTDataFilter::FilterElementsOnDelta( std::shared_ptr<pb::Delta> &delta )
+    {
+        std::vector<pb::Element>                               new_tombstones;
+        std::unordered_map<std::string, ElementFilterCallback> registry_copy;
+        {
+            std::lock_guard lock( element_registry_mutex_ );
```
luizgrz, line 44: maybe use a shared_lock here

src/crdt/impl/crdt_data_filter.cpp
henriqueaklein: I guess it could but not sure if unordered map works by default with regex instead of string, maybe it does, I'll test it

henriqueaklein: well, it doesn't,

PR #134, opened 2025-06-05
dev_migrate_crdt to develop

luizgrz:
## Added
- **MigrationManager** to database migrations on empty nodes
- **Migration test suite** (`migration_sync_test.cpp`) with fixtures for end-to-end validation
- **CMake entries** for tests (`test/src/transaction_sync/CMakeLists.txt`) to include migration tests

## Changed
- **`src/account/CMakeLists.txt`**: added `MigrationManager` sources to `sgns_account`
- **`src/account/GeniusNode.cpp`**: invoke `MigrationManager` during startup
- **`TransactionManager.hpp/.cpp`**: made key methods `public static` and updated logic for transactional migrations
- **`src/crdt/globaldb/globaldb.cpp`**: adjusted initialization to defer to `MigrationManager` for consistent CRDT state

## Fixed
- **`src/storage/rocksdb/rocksdb.cpp`**: detect empty RocksDB to trigger migrations only on new databases

## Removed
- All proof‐verification code paths from the migration workflow

### DIFF
--- a/src/account/MigrationManager.cpp
+++ b/src/account/MigrationManager.cpp
@@ -1,0 +1,80 @@
+/**
+ * @file       MigrationManager.cpp
+ * @brief      Implementation of MigrationManager and migration steps.
+ * @date       2025-05-29
+ * @author     Luiz Guilherme Rizzatto Zucchi
+ * @author     Henrique A. Klein (hklein@gnus.ai)
+ */
+
+#include "account/MigrationManager.hpp"
+#include "account/Migration0_2_0To1_0_0.hpp"
+
+#include <boost/format.hpp>
+#include <boost/system/error_code.hpp>
+#include "base/sgns_version.hpp"
+
+namespace sgns
+{
+
+    std::shared_ptr<MigrationManager> MigrationManager::New(
+        std::shared_ptr<crdt::GlobalDB>                                 newDb,
+        std::shared_ptr<boost::asio::io_context>                        ioContext,
+        std::shared_ptr<ipfs_pubsub::GossipPubSub>                      pubSub,
+        std::shared_ptr<ipfs_lite::ipfs::graphsync::Network>            graphsync,
+        std::shared_ptr<libp2p::protocol::Scheduler>                    scheduler,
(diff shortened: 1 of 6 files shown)


src/account/MigrationManager.hpp
```diff
@@ -0,0 +1,192 @@
...
+#include <ipfs_lite/ipfs/graphsync/impl/network/network.hpp>
+#include <ipfs_lite/ipfs/graphsync/impl/local_requests.hpp>
+#include <libp2p/protocol/common/asio/asio_scheduler.hpp>
+
+namespace sgns
+{
+    /**
+     * @brief Interface for a migration step between two schema versions.
+     */
+    class IMigrationStep
```
henriqueaklein, line 32: Ideally separate once finished, right?

src/account/MigrationManager.cpp
```diff
@@ -0,0 +1,272 @@
...
+        std::shared_ptr<boost::asio::io_context>                        ioContext,
+        std::shared_ptr<ipfs_pubsub::GossipPubSub>                      pubSub,
+        std::shared_ptr<ipfs_lite::ipfs::graphsync::Network>            graphsync,
+        std::shared_ptr<libp2p::protocol::Scheduler>                    scheduler,
+        std::shared_ptr<ipfs_lite::ipfs::graphsync::RequestIdGenerator> generator,
+        const std::string                                              &writeBasePath,
+        const std::string                                              &base58key )
+    {
+        auto instance = std::shared_ptr<MigrationManager>( new MigrationManager() );
+        instance->RegisterStep( std::make_unique<Migration0_2_0To1_0_0>( std::move( newDb ),
```
henriqueaklein, line 30: MigrationManager could be a singleton and not have any dependencies at all I think. Other MigrationSteps could call the singleton->Register. Not sure yet. This is fine for now

src/account/MigrationManager.hpp
luizgrz: Split into interface, migration from 0.2.0 to 1.0.0 and manager. If you dislike the names, just tell me, i cant find any better

PR #149, opened 2025-08-02
Use SGProcessingManager for processing

itsafuu:


src/account/GeniusNode.cpp
```diff
@@ -847,13 +845,11 @@ namespace sgns
 
     outcome::result<std::string> GeniusNode::ProcessImage( const std::string &jsondata )
     {
-        if ( GetTransactionManagerState() != TransactionManager::State::READY )
-        {
-            return outcome::failure( boost::system::error_code{} );
-        }
-        BOOST_OUTCOME_TRYV2( auto &&, CheckProcessValidity( jsondata ) );
```
henriqueaklein: IS there a reason the transaction manager state checking was removed? I think we should keep it

PR #159, opened 2025-08-13
Dev fix nonce consensus

henriqueaklein:
### Fixes

- Implemented nonce consensus via request on libp2p.
- Accounts with the same address sync through each other.
- Hook to processing done and error on GeniusNode changed to protect against async calls (factory design).
- CRDT overwriting criteria changed from string bigger than to string different from.
- Child tokens test fixed (redundant bootstrapping was causing it to be unstable).
- Migration manager fixed, since it won't ever migrate from a dev net ID.
- Signature of the whole transaction and not just the header.
- Protected transaction manager's hooks against async calls.

### Features

- Conflicting transactions being synched.
- Rollback of wrong transactions.
- Highest nonce gathering from the network.
- Mint transactions now rely on the full node confirming them.
- Transactions status inserted (so you can keep track of them)
- Transaction Manager (Genius Node) state inserted, so you can only post transactions when it's ready.
- Generic Account Messenger class that is used for nonce for now, but will be used for other stuff.
- CRDT callbacks changed to a callback registry mechanism, enabling us to stop polling and just insert hooks with a regex to match it.
- CRDT filter changed to remove elements instead of tombstoning them, since this kills the key instead of not replacing it in some cases.
- Many fixes on redundant mutexes.
- Extra logging on some modules, contaning the address of the node.


src/account/AccountMessenger.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+        auto send_ret = SendAccountMessage( envelope, { requests_topic_ } );
+
+        return send_ret;
+    }
+
+    outcome::result<uint64_t> AccountMessenger::GetLatestNonce( uint64_t timeout_ms, uint64_t silent_time_ms )
+    {
+        // Generate a random value
+        std::random_device rd;
```
EduMenges: Usually `random_device` feeds the generator only once because it's `rd()` call is expansive due to it providing a real random value

src/account/AccountMessenger.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+        // Concatenate address and random value
+        std::string to_hash = address_ + std::to_string(random_value);
+
+        // Use HasherImpl to hash the concatenated string
+        sgns::crypto::HasherImpl hasher;
+        auto hash = hasher.sha2_256(gsl::span<const uint8_t>(
+            reinterpret_cast<const uint8_t*>(to_hash.data()), to_hash.size()));
+
+        // Use the first 8 bytes of the hash as req_id
```
EduMenges, line 199: Is this endian dependent?

src/account/AccountMessenger.cpp
```diff
@@ -0,0 +1,427 @@
...
+                        }
+                    }
+                }
+            }
+
+            if ( std::chrono::steady_clock::now() - start_time >= full_timeout )
+            {
+                break; // total timeout reached
+            }
+            std::this_thread::sleep_for( std::chrono::milliseconds( 10 ) );
```
EduMenges, line 249: Why is this sleep here?

src/account/AccountMessenger.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+        logger_->debug( "[{}] Returning highest collected nonce for req_id {}: {}",
+                        address_.substr( 0, 8 ),
+                        req_id,
+                        max_nonce );
+        return max_nonce;
+    }
+
+    outcome::result<void> AccountMessenger::SendAccountMessage( const accountComm::AccountMessage &msg,
+                                                                std::set<std::string>              topics )
```
EduMenges:
```suggestion
                                                                const std::set<std::string>&         topics )
```

src/account/AccountMessenger.cpp
```diff
@@ -0,0 +1,427 @@
...
+        logger_->debug( "[{}] Sending back the nonce {} to {} with req_id {}",
+                        address_.substr( 0, 8 ),
+                        local_nonce,
+                        req.requester_address().substr( 0, 8 ),
+                        resp.request_id() );
+
+        if ( send_ret.has_error() )
+        {
+            logger_->error( "Failed to send NonceResponse" );
+            return;
```
EduMenges:
```suggestion
            logger_->error( "Failed to send NonceResponse" );
```

src/account/GeniusAccount.cpp
```diff
@@ -107,4 +325,109 @@ namespace sgns
...
+        logger_->debug( "Setting local confirmed nonce to {} from {}", nonce, proposed_nonce_ );
+        SetPeerConfirmedNonce( nonce, eth_keypair->GetEntirePubValue() );
+        {
+            std::lock_guard lock( nonce_mutex_ );
+            nonce++;
+            logger_->debug( "Setting the max value between {} and {} as a proposed (next) nonce",
+                            proposed_nonce_,
+                            nonce );
+            proposed_nonce_ = std::max( nonce, proposed_nonce_ );
+        }
```
EduMenges:
```suggestion
        std::lock_guard lock( nonce_mutex_ );
        nonce++;
        logger_->debug( "Setting the max value between {} and {} as a proposed (next) nonce",
                        proposed_nonce_,
                        nonce );
        proposed_nonce_ = std::max( nonce, proposed_nonce_ );
```

src/account/IGeniusTransactions.hpp
```diff
@@ -12,11 +12,10 @@
 #include <string>
 
 #include <boost/format.hpp>
-#include <boost/multiprecision/cpp_int.hpp>
-#include <ProofSystem/EthereumKeyGenerator.hpp>
 
 #include "outcome/outcome.hpp"
 #include "account/proto/SGTransaction.pb.h"
+#include "GeniusAccount.hpp"
```
EduMenges, line 18: The transactions depending on the GeniusAccount is meio paia but faz sentido and there aren't many other options anyway

src/account/IGeniusTransactions.hpp
```diff
@@ -107,9 +111,11 @@ namespace sgns
         }
 
         void FillHash();
+        bool CheckHash();
 
-        std::vector<uint8_t> MakeSignature( std::shared_ptr<ethereum::EthereumKeyGenerator> eth_key );
-        static bool          CheckDAGStructSignature( SGTransaction::DAGStruct dag_st );
+        std::vector<uint8_t> MakeSignature( std::shared_ptr<GeniusAccount> account );
```
EduMenges:
```suggestion
        std::vector<uint8_t> MakeSignature( GeniusAccount& account );
```

src/account/IGeniusTransactions.cpp
```diff
@@ -47,80 +44,53 @@ namespace sgns
...
 
-        nil::marshalling::bincode::field<ecdsa_t::scalar_field_type>::field_element_to_bytes<
-            std::vector<std::uint8_t>::iterator>( std::get<0>( signature ),
-                                                  signed_vector.begin(),
-                                                  signed_vector.begin() + 32 );
-        nil::marshalling::bincode::field<ecdsa_t::scalar_field_type>::field_element_to_bytes<
-            std::vector<std::uint8_t>::iterator>( std::get<1>( signature ),
-                                                  signed_vector.begin() + 32,
-                                                  signed_vector.end() );
+    std::vector<uint8_t> IGeniusTransactions::MakeSignature( std::shared_ptr<GeniusAccount> account )
```
EduMenges:
```suggestion
    std::vector<uint8_t> IGeniusTransactions::MakeSignature( GeniusAccount& account )
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+                account_m->DeleteUTXO( hash );
             }
+
+            m_logger->debug( "[{} - full: {}] Notify {} of deletion of {} to it",
+                             account_m->GetAddress().substr( 0, 8 ),
+                             full_node_m,
+                             dest_infos[i].dest_address,
+                             dest_infos[i].encrypted_amount );
+            topics.emplace( dest_infos[i].dest_address );
         }
```
EduMenges:
```suggestion
        for ( const auto& dest_info: dest_infos )
        {
            if ( dest_info.dest_address == account_m->GetAddress() )
            {
                auto hash = ( base::Hash256::fromReadableString( transfer_tx->dag_st.data_hash() ) ).value();
                account_m->DeleteUTXO( hash );
            }

            m_logger->debug( "[{} - full: {}] Notify {} of deletion of {} to it",
                             account_m->GetAddress().substr( 0, 8 ),
                             full_node_m,
                             dest_info.dest_address,
                             dest_info.encrypted_amount );
            topics.emplace( dest_info.dest_address );
        }
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+        m_logger->debug( "[{} - full: {}] Adding origin address to Broadcast: {}",
+                         account_m->GetAddress().substr( 0, 8 ),
+                         full_node_m,
+                         transfer_tx->GetSrcAddress() );
+        topics.emplace( transfer_tx->GetSrcAddress() );
+
+        m_logger->debug( "[{} - full: {}] Re-parsing inputs to be added as UTXOs",
+                         account_m->GetAddress().substr( 0, 8 ),
+                         full_node_m );
+        for ( auto &input : transfer_tx->GetInputInfos() )
```
EduMenges:
```suggestion
        for ( const auto &input : transfer_tx->GetInputInfos() )
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
         {
-            std::shared_lock<std::shared_mutex> in_lock( incoming_tx_mutex_m );
-            result.reserve( incoming_tx_processed_m.size() );
-            for ( const auto &[key, value] : incoming_tx_processed_m )
+            m_logger->debug( "[{} - full: {}] Fetchin transaction {} ",
```
EduMenges:
```suggestion
            m_logger->debug( "[{} - full: {}] Fetching transaction {} ",
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+            for ( auto i = expected_next_nonce; i < proposed_nonce; ++i )
+            {
+                nonces_to_check.insert( i );
+                m_logger->debug( "[{} - full: {}] Inserting nonce to check: {}",
+                                 account_m->GetAddress().substr( 0, 8 ),
+                                 full_node_m,
+                                 i );
+            }
+
+            CheckTransactionValidity( nonces_to_check );
```
EduMenges: Shouldn't the return of this function be checked?

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+        {
+            m_logger->error( "[{} - full: {}] Local nonce behind - Local: {}, Expected: {}. Waiting to sync",
+                             account_m->GetAddress().substr( 0, 8 ),
+                             full_node_m,
+                             proposed_nonce,
+                             expected_next_nonce );
+        }
+    }
+
+    outcome::result<bool> TransactionManager::CheckTransactionValidity( std::set<uint64_t> nonces_to_check )
```
EduMenges:
```suggestion
    outcome::result<bool> TransactionManager::CheckTransactionValidity( const std::set<uint64_t>& nonces_to_check )
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+                    should_delete = true;
+                    break;
+                }
+                else
+                {
+                    m_logger->debug( "[{} - full: {}] Legacy transaction validated: {}",
+                                     account_m->GetAddress().substr( 0, 8 ),
+                                     full_node_m,
+                                     element.key() );
+                }
```
EduMenges:
```suggestion
                m_logger->debug( "[{} - full: {}] Legacy transaction validated: {}",
                                    account_m->GetAddress().substr( 0, 8 ),
                                    full_node_m,
                                    element.key() );
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+            }
+            m_logger->trace( "[{} - full: {}] Valid proof of {}",
+                             account_m->GetAddress().substr( 0, 8 ),
+                             full_node_m,
+                             element.key() );
+
+            valid_proof = true;
+        } while ( 0 );
+
+        if ( valid_proof == false )
```
EduMenges:
```suggestion
        if ( !valid_proof )
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+
+                    m_logger->debug( "[{} - full: {}] CheckTransactionValidity: Seeing if transaction {} is valid {}",
+                                     account_m->GetAddress().substr( 0, 8 ),
+                                     full_node_m,
+                                     it.second.tx->dag_st.nonce(),
+                                     nonce );
+
+                    if ( it.second.tx->dag_st.nonce() == nonce )
+                    {
+                        bool invalid_tx = false;
```
EduMenges: It would be more readable to start with `valid_tx = true` as reading `!invalid_tx` is obtuse

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+                tombstones.push_back( tx_tombstone );
+            }
+            maybe_tombstones = tombstones;
+        }
+
+        return maybe_tombstones;
+    }
+
+    bool TransactionManager::ShouldReplaceTransaction( const std::shared_ptr<IGeniusTransactions> &existing_tx,
+                                                       const std::shared_ptr<IGeniusTransactions> &new_tx ) const
```
EduMenges:
```suggestion
    bool TransactionManager::ShouldReplaceTransaction( const IGeniusTransactions &existing_tx,
                                                       const IGeniusTransactions &new_tx ) const
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+                             timestamp,
+                             elapsed );
+        }
+
+        return elapsed;
+    }
+
+    int64_t TransactionManager::GetElapsedTime( uint64_t timestamp ) const
+    {
+        return GetElapsedTime( std::move( timestamp ), GetCurrentTimestamp() );
```
EduMenges:
```suggestion
        return GetElapsedTime( timestamp, GetCurrentTimestamp() );
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+    }
+
+    bool TransactionManager::IsTransactionImmutable( const std::shared_ptr<IGeniusTransactions> &tx ) const
```
EduMenges:
```suggestion
    bool TransactionManager::IsTransactionImmutable( const IGeniusTransactions &tx ) const
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+                             transaction_key );
+        }
+        return outcome::success();
+    }
+
+    outcome::result<void> TransactionManager::AddTransactionToProcessedMaps(
+        crdt::CRDTCallbackManager::NewDataPair new_data )
+    {
+        //
+        auto [key, value] = new_data;
```
EduMenges:
```suggestion
        auto [key, value] = new_data;
```

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+    }
+
+    void TransactionManager::ProcessDeletion( std::string key )
+    {
+        m_logger->debug( "[{} - full: {}] Processing deletion of {}",
+                         account_m->GetAddress().substr( 0, 8 ),
+                         full_node_m,
+                         key );
+
+        RemoveTransactionFromProcessedMaps( key );
```
EduMenges: Shouldn't the return of this be checked?

src/account/TransactionManager.cpp
```diff
@@ -906,134 +1279,1208 @@ namespace sgns
...
+                         account_m->GetAddress().substr( 0, 8 ),
+                         full_node_m,
+                         new_data.first );
+
+        m_logger->debug( "[{} - full: {}] Deleting existing entry of key {}",
+                         account_m->GetAddress().substr( 0, 8 ),
+                         full_node_m,
+                         new_data.first );
+
+        AddTransactionToProcessedMaps( new_data );
```
EduMenges: Shouldn't the return of this be checked?

src/crdt/crdt_callback_manager.hpp
```diff
@@ -0,0 +1,82 @@
...
+         * @brief       Executes a registered deleted data callback that matches the key
+         * @param[in]   deleted_key key of the CRDT that was deleted
+         */
+        void DeleteDataCallback( const std::string &deleted_key );
+
+    private:
+        std::shared_mutex           new_data_callback_registry_mutex_;     ///< Mutex to manipulate @ref new_data_callback_registry_
+        NewDataCallbackRegistry     new_data_callback_registry_;           ///< New data callback registry
+        std::shared_mutex           deleted_data_callback_registry_mutex_; ///< Mutex to manipulate @ref deleted_data_callback_registry_
+        DeletedDataCallbackRegistry deleted_data_callback_registry_;       ///< Deleted data callback registry
```
EduMenges, line 79: Would it be better to store the regexes themselves instead of the string pattern? Compiling regexes do take some time.

test/src/transaction_sync/transaction_sync_test.cpp
```diff
@@ -434,29 +439,88 @@ namespace sgns
         sgns::GeniusNode &node          = *node_proc1;
         SendPair( *node_proc1, tx, proof_vect );
 
+        //auto invalid_tx_result_sent = node_proc1->WaitForTransactionOutgoing(
+        //    invalid_tx_id,
+        //    std::chrono::milliseconds( OUTGOING_TIMEOUT_MILLISECONDS ) );
+
+        //std::cout << "Invalid tx went through " << std::endl;
```
EduMenges:
```suggestion
```

test/src/transaction_sync/transaction_sync_test.cpp
```diff
@@ -434,29 +439,88 @@ namespace sgns
...
+        test::assertWaitForCondition(
+            [&]()
+            {
+                return node_proc1->GetTransactionStatus( invalid_tx_id ) ==
+                       TransactionManager::TransactionStatus::VERIFYING;
+            },
+            std::chrono::milliseconds( 20000 ),
+            "Invalid transaction didn't get sent" );
+
+        //EXPECT_EQ( invalid_tx_result_sent, TransactionManager::TransactionStatus::CONFIRMED ); //confirms it's sent
```
EduMenges:
```suggestion
```

test/src/transaction_sync/transaction_sync_test.cpp
```diff
@@ -434,29 +439,88 @@ namespace sgns
...
+
+        std::cout << "wait until its ready" << std::endl;
+
+        EXPECT_EQ( node_proc1->GetBalance(), balance_1_before_invalid ) << "Correct Balance of outgoing transactions";
+        //auto transfer_invalid_received = node_proc2->WaitForTransactionIncoming( invalid_tx_id,
+        //                                                                         std::chrono::milliseconds( 5000 ) );
+
+        //EXPECT_EQ( transfer_invalid_received, TransactionManager::TransactionStatus::INVALID ); //it wasn't received
+
+        // Transfer funds with timeout
```
EduMenges:
```suggestion
```

EduMenges (changes requested):

src/account/AccountMessenger.cpp
henriqueaklein: Otherwise you keep the thread always awake.

src/account/AccountMessenger.cpp
henriqueaklein: Doesn't matter that much I think, it's just to generate a unique-ish ID for a request

src/account/IGeniusTransactions.hpp
henriqueaklein: You could use a forward declaration or a different encapsulation method, but since it's both Genius*something I figured it wouldn't be so bad.

src/account/TransactionManager.cpp
henriqueaklein: Maybe in another method, probably needs some work done in the near future

src/account/TransactionManager.cpp
henriqueaklein: Agreed

src/account/TransactionManager.cpp
henriqueaklein: Right

src/account/TransactionManager.cpp
henriqueaklein: Right

src/crdt/crdt_callback_manager.hpp
henriqueaklein: Not sure because I haven't thought of unordered map works with it. I'd have to see if it's worth it

EduMenges (approved):

PR #179, opened 2025-09-23
Dev crdt traversal fix + dev phase 3.3

henriqueaklein:
This PR contains the merge of dev_phase3.3 with this current branch.

### Changes

- Clang 17 build issues
- Full node topic being processed by non-full nodes
- Rebroadcasting of CRDT heads only when fully processed
- Handling of CID failures
- Duplicated CID request no longer errors out
- Validation of processing results improved
- Marking of bad jobs to prevent trying to reprocess it
- Subtask completion improved


example/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ add_subdirectory(ipfs_pubsub)
 add_subdirectory(account_handling)
 add_subdirectory(processing_mnn)
 add_subdirectory(node_test)
-add_subdirectory(AsyncFile)
+#add_subdirectory(AsyncFile)
```
EduMenges: Just delete it instead of commenting it. You already deleted the related files.

src/account/AccountMessenger.hpp
```diff
@@ -88,9 +88,9 @@ namespace sgns
         std::shared_ptr<ipfs_pubsub::GossipPubSub> pubsub_;             ///< Pubsub instance
 
         /// Future of the subscription to the receiving topic
-        std::future<libp2p::protocol::Subscription> subs_acc_future_;
+        std::shared_future<std::shared_ptr<libp2p::protocol::Subscription>> subs_acc_future_;
         /// Future of the subscription to the requests topic
-        std::future<libp2p::protocol::Subscription> subs_requests_future_;
+        std::shared_future<std::shared_ptr<libp2p::protocol::Subscription>> subs_requests_future_;
```
EduMenges, line 93:
```suggestion
```

src/crdt/impl/crdt_datastore.cpp
```diff
@@ -73,7 +75,85 @@ namespace sgns::crdt
...
+                    while ( dagThreadRunning )
+                    {
+                        if ( auto self = weakptr.lock() )
+                        {
+                            std::unique_lock cvlock( self->dagWorkerMutex_ );
+                            self->dagWorkerCv_.wait_for(
+                                cvlock,
+                                threadSleepTimeInMilliseconds_,
+                                [&] { return !self->rootCIDJobList_.empty() || !dagWorker->dagWorkerThreadRunning_; } );
+                            cvlock.unlock();
```
EduMenges: Won't this unlock cause issues?

src/processing/processing_subtask_queue_accessor_impl.cpp
```diff
@@ -47,6 +54,7 @@ namespace sgns::processing
             m_logger->error( "Tried creating channel with \"RESULT_CHANNEL_ID_{}\" but channel already created",
                              task_id );
         }
+        StartPeriodicStateBroadcast();
```
EduMenges, line 56: Why are you calling this method on the destructor?

src/processing/processing_validation_core.cpp
```diff
@@ -65,6 +65,52 @@ bool ProcessingValidationCore::ValidateResults(
     return areResultsValid;
 }
 
+bool ProcessingValidationCore::ValidateIndividualResult( const SGProcessing::SubTask       &subTask,
+                                                         const SGProcessing::SubTaskResult &result ) const
```
EduMenges: Is this function just returning a bool ok or would it be better to return what actually happened (subtask id, duplicate hashes...)?

EduMenges (changes requested):

src/crdt/impl/crdt_datastore.cpp
henriqueaklein: What type of issues?

src/processing/processing_subtask_queue_accessor_impl.cpp
henriqueaklein: This is called on CreateResultsChannel and not the destructor.

src/processing/processing_validation_core.cpp
henriqueaklein: You are right! On it

src/crdt/impl/crdt_datastore.cpp
EduMenges: This is the expanded snippet:

```cpp
if ( auto self = weakptr.lock() )
{
    std::unique_lock cvlock( self->dagWorkerMutex_ );
    self->dagWorkerCv_.wait_for(
        cvlock,
        threadSleepTimeInMilliseconds_,
        [&] { return !self->rootCIDJobList_.empty() || !dagWorker->dagWorkerThreadRunning_; } );
    cvlock.unlock();
    if ( dagWorker->dagWorkerThreadRunning_ )
    {
        // Pop the job here
        RootCIDJob job_to_process;
        {
            std::unique_lock lock( self->dagWorkerMutex_ );
            if ( self->rootCIDJobList_.empty() )
            {
                continue; // No job, wait again
            }
            job_to_process = self->rootCIDJobList_.front();
            self->rootCIDJobList_.pop();
        }
        // Process the job
        auto process_res = self->ProcessJobIteration( job_to_process );
        if ( process_res.has_failure() )
        {
            self->logger_->error( "CID PROCESSING ERROR: Cleaning up jobs for root CID {}",
                                    job_to_process.root_node_->getCID().toString().value() );
            {
                std::unique_lock       lock( self->dagWorkerMutex_ );
                std::queue<RootCIDJob> temp_queue;
                while ( !self->rootCIDJobList_.empty() )
                {
                    auto job = self->rootCIDJobList_.front();
                    self->rootCIDJobList_.pop();
                    if ( job.root_node_->getCID() != job_to_process.root_node_->getCID() )
                    {
                        temp_queue.push( job );
                    }
                }
                self->rootCIDJobList_ = std::move( temp_queue ); // Restore the filtered queue
            }
            // Cleanup: Delete CID block for the root node
            (void)self->dagSyncer_->DeleteCIDBlock( job_to_process.root_node_->getCID() );
            // Cleanup: Erase from pendingHeadsByRootCID_
            {
                std::lock_guard<std::mutex> lock( self->pendingHeadsMutex_ );
                self->pendingHeadsByRootCID_.erase( job_to_process.root_node_->getCID() );
            }
        }
    }
    else
    {
        dagThreadRunning = false;
    }
}
else
{
    dagThreadRunning = false;
}
```

You are:
1. Unlocking the lock
2. Checking if the thread is still running (is this thread-safe, now that the lock was unlocked?)
3. Reacquiring the lock

This is not a common pattern when using condition variable, usually CVs are for the producer/consumer pattern, to ensure that the consumer are only consuming after the producer has produced. Notice that this may be the intended purpose, it's just not a standard practice.

henriqueaklein: The dagWorkerMutex_ is used to manipulate rootCIDJobList_ so if I keep it locked it's gonna run into a deadlock at some point. There used to be a mutex exclusive to the CV, but is was removed.

henriqueaklein: Fixed it

PR #222, opened 2026-01-08
Add percentage to processing status.

### LINKED ISSUE #221: Update Processing Status to use completion of chunks to represent a percentage
Instead of just DISABLED, IDLE, PROCESSING, if we are in PROCESSING also return complete chunks / total chunks to represent percentage.

itsafuu:
Closes #221

### DIFF
--- a/src/processing/processing_service.cpp
+++ b/src/processing/processing_service.cpp
@@ -406,8 +406,31 @@ namespace sgns::processing
         m_channelListRequestTimeout = channelListRequestTimeout;
     }
 
-    ProcessingServiceImpl::Status ProcessingServiceImpl::GetProcessingStatus() const {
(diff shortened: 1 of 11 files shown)


src/processing/processing_service.cpp
```diff
@@ -406,8 +406,31 @@ namespace sgns::processing
...
+        
+        // Calculate average progress across all processing nodes
+        float totalProgress = 0.0f;
+        size_t nodeCount = 0;
+        for (const auto& [queueId, node] : m_processingNodes) {
+            if (node) {
+                totalProgress += node->GetProgress();
+                ++nodeCount;
+            }
+        }
```
henriqueaklein, line 430: The scope of the m_mutexNodes could be reduced to after the loop, so we don't hold t for math operations

henriqueaklein:
```suggestion
        }
        lock.unlock();
```

henriqueaklein (approved):

PR #224, opened 2026-01-14
Filter rebroadcast

itsafuu:
Only periodically rebroadcast to topics we care about specifically. Request over accountmessenger when we need an updated head list, only every 10 minutes or we find ourselves behind on nonce.


src/crdt/impl/crdt_datastore.cpp
```diff
@@ -976,6 +976,13 @@ namespace sgns::crdt
 
         for ( const auto &[topic_name, cid_set] : head_map ) // Changed from cid_map to head_map
         {
+            // Only broadcast topics that are registered in topicNames_
```
henriqueaklein: I don't think that is correct because the topicNames_ is the topics you are listening to, meanwhile you can broadcast to other topics that are not your own

src/account/TransactionManager.cpp
```diff
@@ -102,6 +102,25 @@ namespace sgns
                 } );
         }
 
+        // Register head request handler to bridge AccountMessenger -> GlobalDB -> CrdtDatastore
+        if ( auto messenger = instance->account_m->GetMessenger() )
```
henriqueaklein: Not sure this is the best place for it, because it seems more of a globaldb thing and not necessarily just transactions. So we have topics which are from processing and whatnot

src/account/TransactionManager.cpp
```diff
@@ -310,6 +344,39 @@ namespace sgns
...
+        auto time_since_last_periodic_sync = std::chrono::duration_cast<std::chrono::minutes>(
+            now - last_periodic_sync_time_ );
+        if ( time_since_last_periodic_sync >= PERIODIC_SYNC_INTERVAL )
+        {
+            m_logger->debug( "[{} - full: {}] Periodic sync - requesting heads after {} minutes",
+                             account_m->GetAddress().substr( 0, 8 ),
+                             full_node_m,
+                             time_since_last_periodic_sync.count() );
+            
+            std::vector<std::string> topics;
```
henriqueaklein: The issue here is that we only request heads from the transaction manager topics, so the processing topic is left out. I think there's also the blockchain topic but I don't recall if I listen to it. Maybe we should keep a record of the topics that sent one someone calls AddListenTopic on globaldb. Actually I think they are the topicNames_ on crdt_datastore.

henriqueaklein (changes requested):

PR #228, opened 2026-01-16
Allowed secure storage to move with directory

EduMenges:


src/account/GeniusAccount.cpp
```diff
@@ -458,17 +484,33 @@ namespace sgns
...
 
-            auto hashed = TW::Hash::sha256( signed_secret );
+            key_seed = nil::crypto3::multiprecision::uint256_t( TW::Hash::sha256( signed_secret ) );
 
-            key_seed = nil::crypto3::multiprecision::uint256_t( hashed );
+            // Create storage if it doesn't exist yet
+            if ( !storage )
+            {
+                ethereum::EthereumKeyGenerator temp_eth_key( key_seed );
+                OUTCOME_TRY( std::vector<uint8_t> vec, base::unhex( temp_eth_key.GetEntirePubValue() ) );
```
henriqueaklein: In the case we created with a previous public key and haven't found the sgns_key, it won't create a new one and will store in the wrong registry.

henriqueaklein (changes requested):

PR #234, opened 2026-02-11
Dev store utxo

EduMenges:


src/account/AccountMessenger.hpp
```diff
@@ -154,7 +154,7 @@ namespace sgns
          * @param[in]   topics Vector of topic names to request heads for
          * @return      outcome::success if request was sent, error otherwise
          */
-        outcome::result<void> RequestHeads( const std::set<std::string> &topics );
+        outcome::result<void> RequestHeads( const std::unordered_set<std::string> &topics );
```
henriqueaklein, line 160: header missing here

src/account/UTXOManager.cpp
```diff
@@ -334,6 +364,121 @@ namespace sgns
         return real_amount == expected_amount && input_amount == params.first.size();
     }
 
+    outcome::result<bool> UTXOManager::LoadUTXOs( std::shared_ptr<crdt::GlobalDB> db )
```
henriqueaklein: The idea is to load/store in rocksdb, not CRDT, because its offline.

src/account/UTXOManager.cpp
```diff
@@ -334,6 +364,121 @@ namespace sgns
...
+                utxos_[address].emplace_back( UTXOState::UTXO_READY,
+                                              GeniusUTXO( hash, utxo.output_idx(), utxo.amount(), token_id ) );
+            }
+        }
+
+        return true;
+    }
+
+    outcome::result<void> UTXOManager::StoreUTXOs( const std::string &address )
+    {
```
henriqueaklein, line 443: Again, store it into rocksdb and not CRDT, so that the interpretation of the UTXOs is a local variable.

src/account/UTXOManager.hpp
```diff
@@ -23,7 +27,8 @@ namespace sgns
             is_full_node_( is_full_node ),
             address_( std::move( address ) ),
             sign_( std::move( sign ) ),
-            verify_signature_( std::move( verify_signature ) )
+            verify_signature_( std::move( verify_signature ) ),
+            db_( nullptr )
```
henriqueaklein: Is there a reason for this nullptr?

src/crdt/globaldb/globaldb.hpp
```diff
@@ -81,14 +81,17 @@ namespace sgns::crdt
          * @param[in] value The value to store.
          * @return outcome::success on success, or outcome::failure otherwise.
          */
-        outcome::result<CID> Put( const HierarchicalKey &key, const Buffer &value, std::set<std::string> topics );
+        outcome::result<CID> Put( const HierarchicalKey                 &key,
+                                  const Buffer                          &value,
+                                  const std::unordered_set<std::string> &topics );
```
henriqueaklein, line 90: Missing unordered_set header

src/crdt/impl/crdt_datastore.cpp
```diff
@@ -645,17 +644,14 @@ namespace sgns::crdt
                         {
                             logger_->error( "{}: Topic {} different from known {} ", __func__, topic, _dontcare );
                         }
-                        std::lock_guard<std::mutex> lock( pendingHeadsMutex_ );
+                        std::lock_guard lock( pendingHeadsMutex_ );
                         pendingHeadsByRootCID_[job.root_node_->getCID()].emplace( cid, topic );
```
henriqueaklein, line 648: This removal might impact on performance. Justin added it because even though the debug doesn't always appears, it enters the method and format stuff and only then it decides not to print it.

src/crdt/crdt_datastore.hpp
```diff
@@ -144,17 +145,17 @@ namespace sgns::crdt
...
          * @param aKey Hierarchical key to put
          * @param aValue Value to be stored
          * @return outcome::success if stored and broadcasted successfully, or outcome::failure otherwise.
          */
-        outcome::result<CID> PutKey( const HierarchicalKey       &aKey,
-                                     const Buffer                &aValue,
-                                     const std::set<std::string> &topics );
+        outcome::result<CID> PutKey( const HierarchicalKey                 &aKey,
+                                     const Buffer                          &aValue,
+                                     const std::unordered_set<std::string> &topics );
```
henriqueaklein, line 159: missing unordered_set header

src/crdt/crdt_heads.hpp
```diff
@@ -104,7 +104,7 @@ namespace sgns::crdt
         * @param topics Topic to get list from
         * @return outcome::failure on error
         */
-        outcome::result<CRDTListResult> GetList( const std::set<std::string> &topics = {} ) const;
+        outcome::result<CRDTListResult> GetList( const std::unordered_set<std::string> &topics = {} ) const;
```
henriqueaklein, line 113: missing unordered_set header

src/account/UTXOManager.cpp
```diff
@@ -334,6 +364,121 @@ namespace sgns
...
+        base::Buffer buf( std::vector<uint8_t>( utxos.ByteSizeLong() ) );
+        if ( !utxos.SerializeToArray( buf.data(), buf.size() ) )
+        {
+            logger_->error( "Failed to serialize to array" );
+            return std::errc::bad_message;
+        }
+
+        crdt::HierarchicalKey key( std::string( DB_PREFIX ) + '/' + address );
+
+        if ( auto result = db_->Put( key, buf, {} ); result.has_error() )
```
henriqueaklein: Just check if overwriting on raw rocksdb works fine

src/account/TransactionManager.hpp
```diff
@@ -205,14 +207,13 @@ namespace sgns
...
+        using TransactionParserFn =
+            outcome::result<void> ( TransactionManager::* )( const std::shared_ptr<IGeniusTransactions> & );
 
-        void                                Update();
-        SGTransaction::DAGStruct            FillDAGStruct( std::string transaction_hash = "" ) const;
-        outcome::result<std::set<uint64_t>> SendTransactionItem( TransactionItem &item );
-        outcome::result<void>               ConfirmTransactions();
-        outcome::result<void>               RollbackTransactions( TransactionItem &item_to_rollback );
+        SGTransaction::DAGStruct                      FillDAGStruct( std::string transaction_hash = "" ) const;
+        outcome::result<std::unordered_set<uint64_t>> SendTransactionItem( TransactionItem &item );
```
henriqueaklein, line 212: missing unordered_set header

src/account/TransactionManager.cpp
```diff
@@ -175,7 +175,7 @@ namespace sgns
 
         if ( !stopped_.load() )
         {
-            CheckTransactions();
+            QueryTransactions();
```
henriqueaklein, line 188: Isn't this query duplicated? Either have it here (after calling utxo manager loadUTXO) or on GeniusNode. I prefer that the LoadUTXO is done here

src/account/TransactionManager.cpp
```diff
@@ -595,24 +595,6 @@ namespace sgns
         return transfer_transaction->GetHash();
     }
```
henriqueaklein, line 653: Isn't ConfirmTransactions necessary anymore?

src/account/UTXOManager.hpp
```diff
@@ -119,20 +124,27 @@ namespace sgns
 
         bool VerifyParameters( const UTXOTxParameters &params, const std::string &address ) const;
 
+        outcome::result<bool> LoadUTXOs( std::shared_ptr<crdt::GlobalDB> db );
+
+        outcome::result<void> StoreUTXOs( const std::string &address );
+
```
henriqueaklein, line 141: The idea is to have a method that returns the current set of UTXOs, so on transaction manager we can iterate it and parse only the ones we want and not query all of them

src/account/TransactionManager.cpp
henriqueaklein: Also, this will only execute if we can't find the UTXO info on the utxo_manager. Ideally here we'd have:

```suggestion
                auto utxo_result = utxo_manager_->LoadUTXOs( globaldb_m );
                if ( utxo_result.has_error() )
                {
                    m_logger->error( "Failed to load UTXOs from storage" );
                }
                if ( utxo_result.has_error() || !utxo_result.value() )
                {
                    m_logger->info( "Loading transactions to mount UTXOs" );
                    QueryTransactions();
                }
                else
                {
                    auto utxo_list = utxo_manager_->GetCurrentUTXOs();
                    for (auto &utxo : utxo_lists)
                    {
                        FetchTransaction(utxo.hash);
                        ParseTransaction....
                        //this should be a ProcessTransaction, which is the method that is called for every transaction on QueryTransactions
                    }
                }
```

src/account/GeniusNode.cpp
```diff
@@ -355,6 +353,18 @@ namespace sgns
                         }
                     } );
                 transaction_manager_->Start();
+
+                auto utxo_result = utxo_manager_.LoadUTXOs( tx_globaldb_ );
```
henriqueaklein: Probably best to do it on TransactionManager

henriqueaklein (changes requested):

src/account/UTXOManager.hpp
EduMenges: Not really

henriqueaklein (approved):

PR #236, opened 2026-02-18
Dev voting impl

henriqueaklein:
### Features
- Per Account Consensus using validator registry.
- Weighted voting using peers on registry.
- Validator registry updates with batches of certificates (default value = 5 certificates per batch)
- Caps on weight of voters according to the role of them
- Registry updates are also ran through consensus to sort out the agreed upon new registry.
- Penalization and decay on Validators.
- Decoupled input validation into InputValidators and expanded chain metadata/replay fields (chain ID, previous hash exposure).
- Certificates using multi sig, preventing hacks.
- UTXO validation without needing to fetch all history. This is done by having a record of UTXOs spent+created in certificates. Peers check their interpretation of certificates to accept/reject proposals.
- Double spending solved with UTXO tracking.
- Added CRDT work journal and integrated it with datastore/filter/callback manager to improve processing of stalled/out-of-order work.
- Migration to database 3.7. Created Migration transaction that depends on the node having tokens on previous versions.
- Switched transaction confirmation to be certificate-based (integrated in TransactionManager with consensus callbacks).

### FIxes
 - Fixed multiple deadlocks/race conditions (consensus thread teardown, datastore worker shutdown, UTXO manager/database shutdown path).
 - Fixed RocksDB reliability issues (retry on open failure, logging improvements, post-migration DB reopen/cleanup behavior).
- Fixed payout to not rely on escrow-release anymore. It's just a transfer now.
- Fixed the posting of the results of a job.
- Fixed the keeping and restoring of UTXOs from memory.
- Fixed zlib link on Windows


cmake/version.cmake
```diff
@@ -1,4 +1,4 @@
-set(PROJECT_VERSION 3.6.0)
+set(PROJECT_VERSION 3.7.0)
```
EduMenges, line 1: Warum?

src/account/EscrowTransaction.hpp
```diff
@@ -13,89 +13,148 @@
...
+        /**
+         * @brief Deserializes a serialized escrow transaction.
+         * @param[in] data Serialized @c SGTransaction::EscrowTx bytes.
+         * @return Shared pointer to the parsed escrow transaction, or nullptr if parsing fails.
+         */
         static std::shared_ptr<EscrowTransaction> DeSerializeByteVector( const std::vector<uint8_t> &data );
 
+        /**
+         * @brief Destroys the escrow transaction.
+         */
```
EduMenges, line 46: Is Doxygen on destructors necessary?

src/account/EscrowTransaction.hpp
```diff
@@ -13,89 +13,148 @@
...
 
-        std::vector<uint8_t> SerializeByteVector() override;
-        uint64_t             GetNumChunks() const;
+        using IGeniusTransactions::SerializeByteVector;
+        /**
+         * @brief Serializes the escrow transaction payload and DAG metadata.
+         * @param[in] dag DAG metadata to serialize into the transaction payload.
+         * @return Serialized @c SGTransaction::EscrowTx bytes.
+         */
+        std::vector<uint8_t> SerializeByteVector( const SGTransaction::DAGStruct &dag ) const override;
```
EduMenges, line 51: Virtual methods already inherit their documentation, right?

src/account/GeniusAccount.cpp
```diff
@@ -157,6 +159,24 @@ namespace
...
+        const auto *end   = input.data() + input.size();
+        auto [ptr, ec]    = std::from_chars( begin, end, value );
+        if ( ec != std::errc() || ptr == begin )
+        {
+            return false;
+        }
+
+        return allow_suffix || ptr == end;
+    }
+
```
EduMenges, line 179: Kinda cringe, but that's C++ for you, I guess

src/account/GeniusNode.cpp
```diff
@@ -516,16 +518,17 @@ namespace sgns
...
+        auto loggerGeniusAccount     = ConfigureLogger( "GeniusAccount", logdir, spdlog::level::err );
+        auto loggerKeyPair           = ConfigureLogger( "KeyPairFileStorage", logdir, spdlog::level::err );
+        auto loggerBlockchain        = ConfigureLogger( "Blockchain", logdir, spdlog::level::err );
+        auto loggerValidator         = ConfigureLogger( "ValidatorRegistry", logdir, spdlog::level::debug );
+        auto loggerProcMgr           = ConfigureLogger( "SGProcessingManager", logdir, spdlog::level::err );
+        auto loggerProcessor         = ConfigureLogger( "SGProcessor", logdir, spdlog::level::err );
+        auto loggerCrdtCallback      = ConfigureLogger( "CRDTCallbackManager", logdir, spdlog::level::err );
+        auto loggerCoinPrices        = ConfigureLogger( "CoinPrices", logdir, spdlog::level::err );
+        auto loggerUTXOManager       = ConfigureLogger( "UTXOManager", logdir, spdlog::level::err );
+        auto loggerConsensusManager  = ConfigureLogger( "ConsensusManager", logdir, spdlog::level::debug );
```
EduMenges, line 531: Are these going to be kept?

src/account/InputValidators.cpp
```diff
@@ -0,0 +1,441 @@
...
+        committed_consumed_payloads.reserve( commitment.consumed_outpoints_size() );
+        for ( const auto &committed_outpoint : commitment.consumed_outpoints() )
+        {
+            auto out_hash_result = base::Hash256::fromSpan(
+                gsl::span( reinterpret_cast<uint8_t *>( const_cast<char *>( committed_outpoint.tx_id_hash().data() ) ),
+                           committed_outpoint.tx_id_hash().size() ) );
+            if ( out_hash_result.has_error() )
+            {
+                return false;
+            }
```
EduMenges, line 153: Why are you doing the hash of the hash?

src/account/InputValidators.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+            const auto outpoint_key = OutPointKey( input.txid_hash_, input.output_idx_ );
+            if ( !seen_inputs.insert( outpoint_key ).second )
+            {
+                return false;
+            }
+            const auto &proof = *proof_it->second;
+
+            const auto &payload = proof.leaf_payload();
+            if ( payload.size() < 32 + 4 + 4 + 32 + 8 )
```
EduMenges: Did you document how the data layout of this paylod is? Also, why didn't you use protobuf?

src/account/InputValidators.cpp
```diff
@@ -0,0 +1,441 @@
...
+            {
+                return false;
+            }
+            const auto payload_output_idx = ReadUInt32BE( reinterpret_cast<const uint8_t *>( payload.data() ) + 32 );
+            if ( payload_output_idx != input.output_idx_ )
+            {
+                return false;
+            }
+            const auto owner_len = ReadUInt32BE( reinterpret_cast<const uint8_t *>( payload.data() ) + 36 );
+            if ( payload.size() < 40 + owner_len + 32 + 8 )
```
EduMenges: Why are you doing these checks if you already done a big one on line 284?

src/account/MigrationAllowList.hpp
```diff
@@ -0,0 +1,122 @@
...
+#include <vector>
+
+class MigrationParamTest;
+
+namespace sgns
+{
+    /**
+     * @brief Stores observed legacy balances and validates migration claim eligibility.
+     */
+    class MigrationAllowList
```
EduMenges, line 28: Is this a good name? It is not quite a list, and does more than simply storage.

src/storage/rocksdb/rocksdb.cpp
```diff
@@ -26,6 +26,7 @@ namespace sgns::storage
 
         // Store a deep copy of options
         l->options_ = std::make_shared<Options>( options );
+        l->logger_  = base::createLogger( "rocksdb" );
```
EduMenges, line 29: But don't we already create a logger in GeniusNode?

test/src/blockchain/CMakeLists.txt
```diff
@@ -1,47 +1,22 @@
...
 
+#addtest(consensus_certificate_test
+#    consensus_certificate_test.cpp
+#)
+#target_link_libraries(consensus_certificate_test
+#    blockchain_genesis
+#    rapidjson
+#    base_crdt_test
+#)
+
```
EduMenges, line 30: Why not deleted it?

src/account/ProcessingTransaction.hpp
```diff
@@ -28,7 +28,8 @@ namespace sgns
 
         ~ProcessingTransaction() override = default;
 
-        std::vector<uint8_t> SerializeByteVector() override;
+        using IGeniusTransactions::SerializeByteVector;
```
EduMenges, line 31: Why is this necessary?

src/account/MigrationTransaction.cpp
```diff
@@ -0,0 +1,203 @@
+/**
+ * @file       MigrationTransaction.cpp
+ * @brief      One-time migration mint transaction implementation.
+ * @date       2026-04-29
+ */
+#include "account/MigrationTransaction.hpp"
```
EduMenges, line 6: One thing that has been bothering me for quite some time now is how we're doing everything under the `account` directory, even tho they don't fit there.

src/account/MigrationTransaction.cpp
```diff
@@ -0,0 +1,203 @@
...
+        }
+        return topics;
+    }
+
+    std::string MigrationTransaction::DeriveUniqueSourceKey( std::string_view from_version,
+                                                             std::string_view source_address,
+                                                             const TokenID   &token_id )
+    {
+        const auto payload =
+            "migration:" + std::string( from_version ) + ":" + std::string( source_address ) + ":" + token_id.ToHex();
```
EduMenges: Conisder using `fmt` instead of string concatenation

src/account/MigrationTransaction.hpp
```diff
@@ -0,0 +1,77 @@
+/**
+ * @file       MigrationTransaction.hpp
+ * @brief      Header file for one-time migration mint transactions.
```
EduMenges: Is this only going to be used for mint migrations? If so, `MintMigrationTransaction` would be more specific and is better future-proof, as we may have migrations for other transactions.

src/account/TransactionManager.cpp
```diff
@@ -32,10 +33,76 @@
...
+    namespace
+    {
+        using utxo_merkle::HashLeaf;
+        using utxo_merkle::HashNode;
+        using utxo_merkle::OutPointKey;
+        using utxo_merkle::ReadUInt32BE;
+        using utxo_merkle::ReadUInt64BE;
+        using utxo_merkle::SerializeUTXOLeafPayload;
+
+        bool ExtractProducedUTXOs( const std::shared_ptr<IGeniusTransactions> &tx, std::vector<GeniusUTXO> &outputs )
```
EduMenges:
```suggestion
        bool ExtractProducedUTXOs( std::shared_ptr<IGeniusTransactions> tx, std::vector<GeniusUTXO> &outputs )
```

src/account/TransactionManager.cpp
```diff
@@ -606,87 +806,264 @@ namespace sgns
...
+        }
+
+        dag.set_previous_hash( previous_hash );
+        dag.set_nonce( nonce );
         dag.set_source_addr( account_m->GetAddress() );
         dag.set_timestamp(
             std::chrono::duration_cast<std::chrono::milliseconds>( timestamp.time_since_epoch() ).count() );
-        dag.set_uncle_hash( "" );
-        dag.set_data_hash( "" ); //filled by transaction class
+        dag.set_uncle_hash( chain_hash );
```
EduMenges, line 826: In the case where `other_chain_hash` is `nullopt`, will the uncle hash be an empty string?

src/account/TransactionManager.cpp
```diff
@@ -954,27 +1281,153 @@ namespace sgns
...
-        return ( this->*it->second.first )( tx );
+        BOOST_OUTCOME_TRY( ( this->*it->second.first )( tx ) );
+        if ( DoesTransactionMutateUTXOState( tx ) && utxo_state_tracking_suppression_.load() == 0 )
+        {
+            UpdateAccountUTXOState( CollectTouchedAccounts( tx ), true );
+        }
+        return outcome::success();
     }
 
     outcome::result<void> TransactionManager::RevertTransaction( const std::shared_ptr<IGeniusTransactions> &tx )
```
EduMenges, line 1297:
```suggestion
    outcome::result<void> TransactionManager::RevertTransaction( std::shared_ptr<IGeniusTransactions> tx )
```

src/account/TransactionManager.cpp
```diff
@@ -954,27 +1281,153 @@ namespace sgns
...
+            state.initialized = true;
+        }
+        state.root = current_root;
+        return state;
+    }
+
+    void TransactionManager::UpdateAccountUTXOState( const std::unordered_set<std::string> &addresses,
+                                                     bool                                   increment_version )
+    {
+        if ( addresses.empty() )
```
EduMenges, line 1395:
```suggestion
        if ( addresses.empty() || !full_node_m )
```
And then you can remove this condition on the loop

src/account/TransactionManager.cpp
```diff
@@ -2093,12 +2627,11 @@ namespace sgns
     {
         for ( const auto &[_, tracked] : tx_processed_m )
         {
-            m_logger->debug( "[{} - full: {}] Searching for hash {}",
-                             account_m->GetAddress().substr( 0, 8 ),
-                             full_node_m,
-                             tx_hash );
-            if ( tracked.tx && tracked.tx->GetHash() == tx_hash &&
-                 tracked.tx->GetSrcAddress() == account_m->GetAddress() )
```
EduMenges, line 2101: Why did you remove the second condition?

src/account/TransactionManager.cpp
```diff
@@ -2665,101 +3033,133 @@ namespace sgns
...
+                    new_tx->GetHash(),
+                    has_cert );
+                return outcome::success();
+            }
+        }
 
-        tx_processed_m[key] = TrackedTx{ new_tx, TransactionStatus::CONFIRMED, nonce };
+        BOOST_OUTCOME_TRY( ChangeTransactionState( new_tx, next_tx_state ) );
 
         return outcome::success();
```
EduMenges:
```suggestion
        return ChangeTransactionState( new_tx, next_tx_state );
```

src/account/TransactionManager.cpp
```diff
@@ -2941,14 +3342,1460 @@ namespace sgns
...
+                                               full_node_m,
+                                               __func__,
+                                               tx.GetHash() );
+            return false;
+        }
+
+        const auto elapsed = GetElapsedTime( ts );
+        if ( elapsed < 0 && timestamp_tolerance_m.count() > 0 &&
+             ( -elapsed ) > static_cast<int64_t>( timestamp_tolerance_m.count() ) )
+        {
```
EduMenges: Why is elapsed negative?

src/account/TransactionManager.cpp
```diff
@@ -2941,14 +3342,1460 @@ namespace sgns
...
+        std::vector<std::vector<uint8_t>> consumed_payloads;
+        consumed_payloads.reserve( inputs.size() );
+        for ( const auto &input : inputs )
+        {
+            auto *committed_input = commitment.add_consumed_outpoints();
+            committed_input->set_tx_id_hash( input.txid_hash_.data(), input.txid_hash_.size() );
+            committed_input->set_output_index( input.output_idx_ );
+
+            std::vector<uint8_t> leaf_payload;
+            leaf_payload.reserve( 32 + 4 );
```
EduMenges: Magic numbers

src/account/TransactionManager.cpp
```diff
@@ -2941,14 +3342,1460 @@ namespace sgns
...
+        leaves.reserve( utxos.size() );
+        for ( const auto &utxo : utxos )
+        {
+            leaves.push_back(
+                { OutPointKey( utxo.GetTxID(), utxo.GetOutputIdx() ), SerializeUTXOLeafPayload( utxo ) } );
+        }
+
+        std::sort( leaves.begin(),
+                   leaves.end(),
+                   []( const SnapshotLeaf &a, const SnapshotLeaf &b ) { return a.payload < b.payload; } );
```
EduMenges, line 4355: Warum?

src/account/BridgeConsensusAdapter.cpp
```diff
@@ -0,0 +1,102 @@
...
+        return ConsensusManager::CreateGenericSubject(
+            account_id,
+            std::string( kBridgeEventSubjectType ),
+            eth::bridge_event_claim_payload( claim ) );
+    }
+
+    outcome::result<eth::BridgeEventClaim> DecodeBridgeEventConsensusSubject(
+        const ConsensusManager::Subject &subject )
+    {
+        if ( !ConsensusManager::SubjectTypeMatches( subject, std::string( kBridgeEventSubjectType ) ) )
```
EduMenges, line 50: Sus

src/account/BridgeConsensusAdapter.cpp
```diff
@@ -0,0 +1,102 @@
...
+        {
+            return outcome::failure( std::errc::invalid_argument );
+        }
+        if ( !CheckOpaqueSubjectPayload( subject ) )
+        {
+            return outcome::failure( std::errc::invalid_argument );
+        }
```
EduMenges, line 57: Why not add an `||` statement to the previous `if`? They're returning the same thing.

src/account/BridgeConsensusAdapter.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+        const auto payload = std::vector<uint8_t>( subject.payload().begin(), subject.payload().end() );
+        auto       claim   = eth::decode_bridge_event_claim_payload( payload );
+        if ( !claim.has_value() )
+        {
+            return outcome::failure( std::errc::invalid_argument );
+        }
+        return claim.value();
```
EduMenges, line 65:
```suggestion
        auto       claim   = eth::decode_bridge_event_claim_payload( payload );
        return claim;
```

src/account/InputValidators.cpp
```diff
@@ -0,0 +1,445 @@
...
+        }
+
+        std::vector<uint8_t> SerializeOutputLeafPayload( const base::Hash256 &txid_hash,
+                                                         uint32_t             output_index,
+                                                         const std::string   &owner_address,
+                                                         gsl::span<const uint8_t> token_bytes,
+                                                         uint64_t                 amount )
+        {
+            std::vector<uint8_t> payload;
+            payload.reserve( 32 + 4 + 4 + owner_address.size() + token_bytes.size() + 8 );
```
EduMenges: Magic numbers

src/account/UTXOManager.cpp
```diff
@@ -9,6 +12,68 @@
 
 namespace sgns
 {
+    namespace
+    {
+        void RemoveOutPointFromVector( std::vector<OutPoint> &outpoints, const OutPoint &target )
+        {
+            outpoints.erase( std::remove( outpoints.begin(), outpoints.end(), target ), outpoints.end() );
+        }
```
EduMenges: Is this method necessary? It could be inlined

src/blockchain/Consensus.hpp
```diff
@@ -0,0 +1,692 @@
...
+#include "crdt/globaldb/globaldb.hpp"
+#include "crdt/proto/delta.pb.h"
+#include "ipfs_pubsub/gossip_pubsub.hpp"
+#include "outcome/outcome.hpp"
+
+namespace sgns
+{
+    inline constexpr const char *kNonceSubjectType         = "sgns.nonce.v1";
+    inline constexpr const char *kTaskResultSubjectType    = "sgns.task_result.v1";
+    inline constexpr const char *kRegistryBatchSubjectType = "sgns.registry_batch.v1";
```
EduMenges:
```suggestion
    constexpr std::string_view kNonceSubjectType         = "sgns.nonce.v1";
    constexpr std::string_view kTaskResultSubjectType    = "sgns.task_result.v1";
    constexpr std::string_view kRegistryBatchSubjectType = "sgns.registry_batch.v1";
```
Also use THIS_CASE for constants since it is the default of the project.

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+        if ( !signer )
+        {
+            ConsensusManagerLogger()->error( "{}: Failed to create ConsensusManager: signer is null", __func__ );
+            return nullptr;
+        }
+        if ( address.empty() )
+        {
+            ConsensusManagerLogger()->error( "{}: Failed to create ConsensusManager: address is empty", __func__ );
+            return nullptr;
+        }
```
EduMenges, line 64: Are these checks necessary? I think most of them were already validated by whomever creates the `ConsensusManager` anyways

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+        if ( timestamp_ms == 0 )
+        {
+            return false;
+        }
+        const auto now_ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                                std::chrono::system_clock::now().time_since_epoch() )
+                                .count();
+        const auto window_ms = timestamp_window_.count();
+        const auto ts_ms     = static_cast<std::int64_t>( timestamp_ms );
+        return ( ts_ms >= now_ms - window_ms ) && ( ts_ms <= now_ms + window_ms );
```
EduMenges: Can't this be done without converting the duration to integer?

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+        }
+
+        sgns::crypto::HasherImpl hasher;
+        auto                     hash = hasher.sha2_256(
+            gsl::span<const uint8_t>( reinterpret_cast<const uint8_t *>( proposal.proposal_id().data() ),
+                                      proposal.proposal_id().size() ) );
+        uint64_t base_index = 0;
+        for ( size_t i = 0; i < sizeof( uint64_t ) && i < hash.size(); ++i )
+        {
+            base_index = ( base_index << 8 ) | hash[i];
```
EduMenges, line 418: :eyes:

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+                __func__,
+                GetPrintableSubjectHash( proposal.subject() ),
+                proposal.registry_epoch(),
+                registry.epoch() );
+            return outcome::failure( std::errc::invalid_argument );
+        }
+
+        uint64_t              total_weight    = ValidatorRegistry::TotalWeight( registry );
+        uint64_t              approved_weight = 0;
+        std::set<std::string> seen;
```
EduMenges:
```suggestion
        std::unordered_set<std::string> seen;
```

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+        if ( !CheckProposal( proposal ) )
+        {
+            ConsensusManagerLogger()->error( "{}: rejected: Invalid proposal for hash {} proposal_id={}",
+                                             __func__,
+                                             GetPrintableSubjectHash( proposal.subject() ),
+                                             proposal.proposal_id().substr( 0, 8 ) );
+            return;
+        }
+
+        if ( !IsTimestampSane( proposal.timestamp() ) )
```
EduMenges, line 1147: Why isn't this check inside `CheckProposal`?

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+
+        ConsensusManagerLogger()->debug( "{}: certificate accepted key={}", __func__, element.key() );
+        return std::nullopt;
+    }
+
+    void ConsensusManager::CertificateReceived( crdt::CRDTCallbackManager::NewDataPair new_data,
+                                                const std::string                     &cid )
+    {
+        auto [key, value] = new_data;
+        (void)cid;
```
EduMenges, line 1567: ?

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+            return BestHash( curr_hash, cand_hash ) == cand_hash;
+        }
+
+        return candidate.proposal_id() < current.proposal_id();
+    }
+
+    const std::string &ConsensusManager::BestHash( const std::string &a, const std::string &b )
+    {
+        return ( a <= b ) ? a : b;
+    }
```
EduMenges, line 2089: ?

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+        if ( !subject.SerializeToString( &serialized ) )
+        {
+            ConsensusManagerLogger()->error( "{}: failed: serialization error", __func__ );
+            return outcome::failure( std::errc::invalid_argument );
+        }
+
+        sgns::crypto::HasherImpl hasher;
+        auto                     hash = hasher.sha2_256(
+            gsl::span<const uint8_t>( reinterpret_cast<const uint8_t *>( serialized.data() ), serialized.size() ) );
+        ConsensusManagerLogger()->debug( "{}: success", __func__ );
```
EduMenges: This pattern repeats a lot, maybe we could have hasher having an overload for `const void*` and size

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+        {
+            return outcome::failure( std::errc::invalid_argument );
+        }
+        return payload;
+    }
+
+    bool ConsensusManager::SetSubjectTypeHash( Subject *subject )
+    {
+        return subject != nullptr && subject->has_subject_type_hash() && !subject->subject_type_hash().hash().empty();
+    }
```
EduMenges: Why "Set" if it is not setting anything?

src/blockchain/Consensus.cpp
```diff
@@ -0,0 +1,2763 @@
...
+                return false;
+            }
+            if ( payload.value().batch_root().empty() )
+            {
+                ConsensusManagerLogger()->error( "{}: subject registry_batch batch_root is empty", __func__ );
+                return false;
+            }
+            return true;
+        }
+
```
EduMenges:
```suggestion
        }

```

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -184,64 +236,64 @@ namespace sgns::blockchain
...
                 break;
         }
 
-        if ( multiplier == 0 )
+        if ( weight == 0 )
         {
-            logger_->debug( "{}: multiplier is zero, weight=0", __func__ );
+            logger_->debug( "{}: weight is zero", __func__ );
             return 0;
         }
```
EduMenges, line 282: In which situations will this occur?

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
-    outcome::result<std::vector<uint8_t>> ValidatorRegistry::ComputeUpdateSigningBytes(
-        const RegistryUpdate &update ) const
+    std::string ValidatorRegistry::BuildBatchKey( const std::string &base_registry_cid, uint64_t base_registry_epoch )
     {
-        logger_->trace( "{}: entry validators={}", __func__, update.registry().validators().size() );
-        validator::RegistrySigningPayload payload;
-        *payload.mutable_registry() = update.registry();
-        payload.set_prev_registry_hash( update.prev_registry_hash() );
+        return base_registry_cid + ":" + std::to_string( base_registry_epoch );
+    }
```
EduMenges: Could be inlined

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
-        return std::vector<uint8_t>( serialized.begin(), serialized.end() );
+        std::string payload;
+        for ( size_t i = 0; i < subject_hashes.size(); ++i )
+        {
+            if ( i > 0 )
+            {
+                payload.push_back( '\n' );
+            }
+            payload += subject_hashes[i];
+        }
```
EduMenges:
```suggestion
        payload += subject_hashes[0];
        for ( size_t i = 1; i < subject_hashes.size(); ++i )
        {
            payload.push_back( '\n' );
            payload += subject_hashes[i];
        }
```

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
+            {
+                return outcome::success();
+            }
+            pending_batch_subject_ids_.insert( batch_hash_result.value() );
+        }
+
+        return submitter( subject_result.value() );
+    }
+
+    outcome::result<ValidatorRegistry::BatchSubjectDecision> ValidatorRegistry::EvaluateBatchSubject(
```
EduMenges, line 859: Why have an outcome as the return if you're only returning success on all ocasions?

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
+        {
+            std::lock_guard<std::mutex> lock( batch_mutex_ );
+            if ( finalized_batch_subject_ids_.find( subject_hash ) != finalized_batch_subject_ids_.end() )
+            {
+                return BatchCertificateDecision::Approve;
+            }
+            if ( applying_batch_subject_ids_.find( subject_hash ) != applying_batch_subject_ids_.end() )
+            {
+                return BatchCertificateDecision::Approve;
+            }
```
EduMenges:
```suggestion
            if ( finalized_batch_subject_ids_.find( subject_hash ) != finalized_batch_subject_ids_.end() 
                  || applying_batch_subject_ids_.find( subject_hash ) != applying_batch_subject_ids_.end() )
            {
                return BatchCertificateDecision::Approve;
            }
```

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
+            std::lock_guard<std::mutex> lock( batch_mutex_ );
+            applying_batch_subject_ids_.erase( subject_hash );
+            return BatchCertificateDecision::Reject;
+        }
+        if ( !ValidateCertificate( certificate, current_registry_result.value() ) )
+        {
+            std::lock_guard<std::mutex> lock( batch_mutex_ );
+            applying_batch_subject_ids_.erase( subject_hash );
+            return BatchCertificateDecision::Reject;
+        }
```
EduMenges, line 917:
```suggestion
        auto current_registry_result = LoadRegistry();
        if ( current_registry_result.has_error() || !ValidateCertificate( certificate, current_registry_result.value() ) )
        {
            std::lock_guard<std::mutex> lock( batch_mutex_ );
            applying_batch_subject_ids_.erase( subject_hash );
            return BatchCertificateDecision::Reject;
        }
```

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
+                    self->logger_->error( "{}: failed storing batch registry update subject_hash={} error={}",
+                                          __func__,
+                                          subject_hash.substr( 0, 8 ),
+                                          store_result.error().message() );
+                    return;
+                }
+                self->pending_batch_subject_ids_.erase( subject_hash );
+                self->finalized_batch_subject_ids_.insert( subject_hash );
+            } )
+            .detach();
```
EduMenges, line 1026: Why is this thread necessary?

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
+            return outcome::success( std::optional<uint64_t>{} );
+        }
+
+        const auto *validator = FindValidator( cached_registry_.value(), validator_id );
+        if ( !validator || validator->status() != Status::ACTIVE )
+        {
+            return outcome::success( std::optional<uint64_t>{} );
+        }
+
+        return outcome::success( std::optional<uint64_t>{ validator->weight() } );
```
EduMenges, line 1045: Why return `success` everywhere?

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -456,204 +544,860 @@ namespace sgns::blockchain
...
+                current_id = cached_registry_id_;
+            }
+            if ( current_id.empty() || prev_registry_cid != current_id )
+            {
+                logger_->error( "{}: prev registry CID mismatch", __func__ );
+                return false;
+            }
+
+            logger_->info( "{}: certificate-based update verified", __func__ );
+            return true;
```
EduMenges, line 1342: Since the else branch returns at the end anyways, invert the `if` condition, put everything in the else block there, and do an early return.

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -678,19 +1422,570 @@ namespace sgns::blockchain
...
+        }
+
+        if ( !IsQuorum( approved_weight, total_weight ) )
+        {
+            logger_->error( "{}: quorum not reached approved={} total={}", __func__, approved_weight, total_weight );
+            result.approved.clear();
+            result.unregistered.clear();
+            result.registered_votes.clear();
+            result.unregistered_votes.clear();
+            return result;
```
EduMenges: Why clear the result instead of just returning a new, empty result?

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -678,19 +1422,570 @@ namespace sgns::blockchain
...
+        std::unordered_set<std::string> participants;
+        participants.reserve( registered_votes.size() + unregistered_votes.size() );
+        for ( const auto &pair : registered_votes )
+        {
+            participants.insert( pair.first );
+        }
+        for ( const auto &pair : unregistered_votes )
+        {
+            participants.insert( pair.first );
+        }
```
EduMenges, line 1652: Kinda goofy, but there are not many ways to make it better

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -678,19 +1422,570 @@ namespace sgns::blockchain
...
+        ApplyInactivityDecay( entries, participants );
+        ApplyTotalWeightCap( entries );
+
+        std::sort( entries.begin(),
+                   entries.end(),
+                   []( const ValidatorEntry &a, const ValidatorEntry &b )
+                   { return a.validator_id() < b.validator_id(); } );
```
EduMenges, line 1659: Why are you doing these sortings? You also do them on other places, what do they serve?

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -678,19 +1422,570 @@ namespace sgns::blockchain
...
+            if ( entry.status() != Status::ACTIVE )
+            {
+                continue;
+            }
+            if ( participants.find( entry.validator_id() ) != participants.end() )
+            {
+                continue;
+            }
+            uint32_t missed = static_cast<uint32_t>( entry.missed_epochs() );
+            if ( missed < std::numeric_limits<uint32_t>::max() )
```
EduMenges, line 1823: Why?

src/crdt/impl/crdt_work_journal.cpp
```diff
@@ -0,0 +1,326 @@
...
+        auto                        maybe_entry = GetEntryUnlocked( key );
+        if ( !maybe_entry.has_value() )
+        {
+            return;
+        }
+
+        auto entry            = maybe_entry.value();
+        entry.state           = State::Stalled;
+        entry.updated_at_ms   = NowMs();
+        entry.lease_until_ms  = entry.updated_at_ms + static_cast<uint64_t>( std::max<int64_t>( 0, lease.count() ) );
```
EduMenges, line 87: Why are you taking the max out of 0 and lease? Can lease be negative?

src/crdt/impl/crdt_work_journal.cpp
```diff
@@ -0,0 +1,326 @@
...
+            return std::nullopt;
+        }
+        entry.key = key_str.substr( pos + std::string( "/crdt/work/" ).size() );
+        return entry;
+    }
+
+    std::string CRDTWorkJournal::SerializeEntry( const Entry &entry )
+    {
+        return "v1|" + std::to_string( static_cast<int>( entry.state ) ) + "|" + std::to_string( entry.attempt_count ) +
+               "|" + std::to_string( entry.updated_at_ms ) + "|" + std::to_string( entry.lease_until_ms );
```
EduMenges, line 277: Why do this instead of using protobuf?

src/crdt/impl/crdt_work_journal.cpp
```diff
@@ -0,0 +1,326 @@
...
+            if ( pos == std::string::npos )
+            {
+                out.push_back( value.substr( start ) );
+                break;
+            }
+            out.push_back( value.substr( start, pos - start ) );
+            start = pos + 1;
+        }
+        return out;
+    }
```
EduMenges, line 296: Could this be inlined? it seems to be used on one place only

src/crdt/impl/crdt_work_journal.cpp
```diff
@@ -0,0 +1,326 @@
...
+        }
+        base::Buffer key_buf;
+        key_buf.put( BuildStorageKey( key ) );
+        auto maybe_value = datastore_->get( key_buf );
+        if ( maybe_value.has_error() )
+        {
+            return std::nullopt;
+        }
+        return DeserializeEntry( BuildStorageKey( key ), maybe_value.value().toString() );
+    }
```
EduMenges, line 312: Do we need to have a mutex that guards the datastore? I think rocksdb can already handle this kind of synchronization, right?

test/src/blockchain/CMakeLists.txt
```diff
@@ -1,47 +1,38 @@
...
-    blockchain_genesis_test.cpp
-    )
+#addtest(consensus_certificate_test
+#    consensus_certificate_test.cpp
+#)
+#target_link_libraries(consensus_certificate_test
+#    blockchain_genesis
+#    rapidjson
+#    base_crdt_test
+#)
```
EduMenges, line 29: ?

test/src/multiaccount/multi_account_sync.cpp
```diff
@@ -18,16 +19,23 @@
...
 #include <random>
 #include <ctime>
+#include <tuple>
 
 #include <boost/program_options.hpp>
 #include <boost/format.hpp>
 #include <boost/asio.hpp>
 #include "local_secure_storage/impl/json/JSONSecureStorage.hpp"
+#define private public
+#define protected public
```
EduMenges: lmao

test/src/multiaccount/multi_account_sync.cpp
```diff
@@ -136,7 +134,7 @@ class MultiAccountTest : public ::testing::Test
     }
 };
 
-TEST_F( MultiAccountTest, SyncThroughEachOther )
+TEST_F( MultiAccountTest, DISABLED_SyncThroughEachOther )
```
EduMenges: Why disable?

test/testutil/mint_source_hash.hpp
```diff
@@ -0,0 +1,61 @@
...
+                                                  std::chrono::high_resolution_clock::now().time_since_epoch().count() );
+                                              std::random_device rd;
+                                              std::seed_seq      seed{ static_cast<uint32_t>( now ),
+                                                                  static_cast<uint32_t>( now >> 32 ),
+                                                                  rd(),
+                                                                  rd(),
+                                                                  rd(),
+                                                                  rd() };
+                                              return std::mt19937_64( seed );
+                                          }() );
```
EduMenges, line 39: Why?

EduMenges (changes requested):

EduMenges (changes requested):
Mostly minor things since I can't do a high quality review on this one

cmake/version.cmake
henriqueaklein: The validator registry was changed and consensus+finalization was created. Version 3.6 didn't have that, so we need to get funds from 3.6, create a certificate for a migration transaction and post it on 3.7

src/account/EscrowTransaction.hpp
henriqueaklein: If it's trivial, no, but if it does extra work and you'd like to detail how it goes about deleting itself it's not a bad idea. On this case it's basically a batter of symmetry

src/account/GeniusNode.cpp
henriqueaklein: What do you mean? the log levels?

src/account/InputValidators.cpp
henriqueaklein: It's just a deserialization of a hash into out_hash_result

src/account/EscrowTransaction.hpp
henriqueaklein: They do by default, yes

src/account/InputValidators.cpp
henriqueaklein: I didn't document it no. Protobuf is deterministic in a version but can't be relied for canonical serialization between versions. This is more lightweight and better as far as compatibility goes.

src/account/InputValidators.cpp
henriqueaklein: This one checks "owner_len", so they are not the same.

src/account/MigrationAllowList.hpp
henriqueaklein: it's actually an allow list with extra info. It stores peers who are allowed migration and their balances. I'm open to any good suggestions for a name

src/storage/rocksdb/rocksdb.cpp
henriqueaklein: The logger component doesn't work like that. From what I can remember it's a pointer for a singleton. That's why multiple classes have the logger on their headers as well as the GeniusNode.

test/src/blockchain/CMakeLists.txt
henriqueaklein: I don't want to delete it, I want to fix it. If i remove it I'll forget to do it

src/account/ProcessingTransaction.hpp
henriqueaklein: To enable SerializeByteVector from the base class to be called by a derived instance.

src/account/MigrationTransaction.cpp
henriqueaklein: You're not wrong. The structure needs fixing, but this PR is already too big

src/account/MigrationTransaction.hpp
henriqueaklein: Actually there's a misconception on how you understood that description. The MigrationTransaction is a special Mint, that gathers all old transactions (escrow release, mint, transfers), sums it up and submit into consensus. This is then voted by the genesis node using the migration allow list. But I get how it can be misleading. I'm going to rewrite the description.

src/account/TransactionManager.cpp
henriqueaklein: passing by value is useful when the method/class wants to store ownership as well. If the method only observes/read from it there's no point

src/account/TransactionManager.cpp
henriqueaklein: Yes, it's not used if that is the case.

src/account/TransactionManager.cpp
henriqueaklein: No point in pass by value if method doesn't want ownership

src/account/TransactionManager.cpp
henriqueaklein: Close, but it actually run if not the full node and the addresses have account_m->GetAddress()

src/account/TransactionManager.cpp
henriqueaklein: To enable grabbing transactions on the map that aren't posted by self

src/account/TransactionManager.cpp
henriqueaklein: Because that's an error. good catch

src/account/TransactionManager.cpp
henriqueaklein: to guarantee canonical order when computing the witness, as the merkle root changes if you change the order.

src/account/BridgeConsensusAdapter.cpp
henriqueaklein: I don't know if that will be used in the end

src/account/BridgeConsensusAdapter.cpp
henriqueaklein: Maybe to insert a different error down the road.

src/account/BridgeConsensusAdapter.cpp
henriqueaklein: decode_bridge_event_claim_payload returns an optional, not outcome.

src/account/UTXOManager.cpp
henriqueaklein: Necessary is a strong word for it, but agreed, removing it now

src/blockchain/Consensus.cpp
henriqueaklein: Encapsulation is key, even if we verify outside this way we are guaranteed to detect misuse and log it out

src/blockchain/Consensus.cpp
henriqueaklein: It returns the best hash

src/blockchain/Consensus.cpp
henriqueaklein: Because timestamp checking is only done at the time of processing the data. On this case the HandleProposal is called when consensus is running and needs to check if the timestamp is consistent. CheckProposal verifies the integrity of the proposal, so it can be called anytime we want to validated it afterwards.

src/crdt/impl/crdt_work_journal.cpp
henriqueaklein: Yes, rocksdb does, but the other variables on this class don't have guards against it. It's needed to mark the states correctly.

test/src/blockchain/CMakeLists.txt
henriqueaklein: Gonna work on that later, so I won't delete or else I'll forget it

test/src/multiaccount/multi_account_sync.cpp
henriqueaklein: Well, I forgot I disabled it since it wasn't working when I did, I should enable and test it. good catch

test/src/multiaccount/multi_account_sync.cpp
henriqueaklein: yes, that is terrible. It's a hack to access private members of a class, which I personally thought it wasn't that bad until Windows CI blew up because of this in another test.

src/account/TransactionManager.cpp
henriqueaklein: That being said, const reference without the pointer is better

src/blockchain/Consensus.cpp
henriqueaklein: what is your question?

src/blockchain/Consensus.cpp
henriqueaklein: Good question. Renaming that

src/blockchain/ValidatorRegistry.cpp
henriqueaklein: On a misuse

src/blockchain/ValidatorRegistry.cpp
henriqueaklein: the lambda which calls it does return an outcome, so it would be easier. The prototype was created thinking we might insert errors at some point

src/blockchain/ValidatorRegistry.cpp
henriqueaklein: otherwise we end up writing on the database on the same thread that writes on the database, resulting in a deadlock.

src/blockchain/ValidatorRegistry.cpp
henriqueaklein: the error is reserved in case we want to get the validator from another place. Basically a while ago we thought of making it possible to fetch validators from other places and not just our network and that would result in errors. By leaving the outcome return we already treat this for future implementation

src/blockchain/ValidatorRegistry.cpp
henriqueaklein: what condition?

src/blockchain/ValidatorRegistry.cpp
henriqueaklein: Canonical ordering in all peers that run the same code.

src/blockchain/ValidatorRegistry.cpp
henriqueaklein: so it doesn't overflow

src/crdt/impl/crdt_work_journal.cpp
henriqueaklein: Yes, lease can be negative.

src/crdt/impl/crdt_work_journal.cpp
henriqueaklein: that's the key, not the payload.

src/crdt/impl/crdt_work_journal.cpp
henriqueaklein: It could, but it won't 😃

src/account/GeniusNode.cpp
EduMenges: Yes

src/crdt/impl/crdt_work_journal.cpp
EduMenges: Which variable? I only see these in the header file:

```cpp
        std::shared_ptr<storage::rocksdb> datastore_; ///< Backing datastore for journal state.
        mutable std::mutex                mutex_;     ///< Synchronizes public journal operations.
```

test/src/multiaccount/multi_account_sync.cpp
EduMenges: But why are they there in the first place?

src/account/InputValidators.cpp
```diff
@@ -284,37 +286,43 @@ namespace sgns
...
+            if ( payload.size() <
+                 OWNER_ADDRESS_OFFSET + TOKEN_ID_BYTES_IN_PAYLOAD + AMOUNT_BYTES_IN_PAYLOAD )
             {
                 return false;
             }
 
             auto payload_hash_result = base::Hash256::fromSpan(
-                gsl::span( reinterpret_cast<uint8_t *>( const_cast<char *>( payload.data() ) ), 32 ) );
+                gsl::span( reinterpret_cast<uint8_t *>( const_cast<char *>( payload.data() ) ), HASH256_BYTES ) );
             if ( payload_hash_result.has_error() || payload_hash_result.value() != input.txid_hash_ )
```
EduMenges, line 297: Can you use the new pointer interface here?

EduMenges (approved):

src/account/InputValidators.cpp
henriqueaklein: that's not the sha method.

test/src/multiaccount/multi_account_sync.cpp
henriqueaklein: To access private members of the class. I already removed them

src/crdt/impl/crdt_work_journal.cpp
henriqueaklein: the key itself. What happens if you call MarkSeen and in the middle of it you call MarkDone? What is the value without and mutex?

PR #238, opened 2026-02-27
Dev fetch utxos

henriqueaklein:
The current state of the develop branch fails on already initialized databases that have UTXOs, because the parsing is not done in all transactions and the nonce grabbing checks if the transaction is in the in-memory structure, which not all of them will be. So these changes were implemented:

- InitNonce removed and replaced with InitTransactions:
    - Doesn't check on nonces from the network and instead check if we have all the UTXOs we need to start transactioning
    - Relies on local UTXOs. If no local UTXOs found, it's gonna request from the network.
    - An improvement should be to fetch the UTXOs form the network regardless and try to see which one is more updated.
- Request UTXOs on pubsub
    - Pubsub messages that request a list of UTXOs from other peers.
    - If we are outdated, we could grab the most trusted information from other peers (taking into account validators and soft consensus)
    - Done on initialization instead of requesting for nonces.
- Request Transactions on pubsub
    - We can request a transaction by hash on pubsub.
    - The grabbing of the content is still done by graphsync.
    - This is means to avoid having to go through all the CRDT graph (although this wasn't removed yet)
    - In order to allow this, CID info is being stored on local rocksdb as a look up table for transactions.
- Nonces stores in rocksdb.
    - With not all transactions being parsed anymore we need the latest nonce stored somewhere.
    - We record the confirmed nonce peers we receive data from to avoid re-parsing everything.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -564,7 +564,7 @@ jobs:
 
       - name: Run tests (Linux)
         working-directory: ${{ github.workspace }}/SuperGenius/${{ env.BUILD_DIRECTORY }}
-        if: ${{ matrix.build-type == 'Release' && matrix.target == 'Linux' }}
+        if: ${{ matrix.target == 'Linux' }}
         shell: bash
         run: |
           dbus-run-session -- bash -c '\
@@ -575,4 +575,4 @@ jobs:
             lock-on-idle=false
             lock-after=false" > ~/.local/share/keyrings/login.keyring && \
             eval $(printf "\n" | gnome-keyring-daemon --unlock --components=secrets) && \
-            ctest . -j -C ${{ matrix.build-type }}'
+            ctest . --verbose -j -C ${{ matrix.build-type }}'
--- a/src/account/AccountMessenger.cpp
+++ b/src/account/AccountMessenger.cpp
@@ -9,6 +9,7 @@
 #include <boost/format.hpp>
 #include <future>
 #include <algorithm>
+#include <type_traits>
 #include "AccountMessenger.hpp"
 #include "base/sgns_version.hpp"
 #include "crypto/hasher/hasher_impl.hpp"
@@ -212,6 +213,9 @@ namespace sgns
                 case accountComm::AccountMessage::kBlockCidRequest:
                     HandleBlockCidRequest( acc_msg.block_cid_request() );
                     break;
+                case accountComm::AccountMessage::kTransactionRequest:
+                    HandleTransactionRequest( acc_msg.transaction_request() );
+                    break;
                 case accountComm::AccountMessage::kUtxoRequest:
                     HandleUTXORequest( acc_msg.utxo_request() );
                     break;
@@ -255,6 +259,9 @@ namespace sgns
                     break;
                 case accountComm::AccountMessage::kUtxoResponse:
                     HandleUTXOResponse( acc_msg.utxo_response() );
+                    break;
+                case accountComm::AccountMessage::kTransactionRequest:
+                    logger_->error( "{}: Unexpected response received ", __func__ );
                     break;
                 case accountComm::AccountMessage::kUtxoRequest:
                     logger_->error( "{}: Unexpected response received ", __func__ );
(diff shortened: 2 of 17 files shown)


.github/workflows/cmake.yml
```diff
@@ -575,4 +575,4 @@ jobs:
             lock-on-idle=false
             lock-after=false" > ~/.local/share/keyrings/login.keyring && \
             eval $(printf "\n" | gnome-keyring-daemon --unlock --components=secrets) && \
-            ctest . -j -C ${{ matrix.build-type }}'
+            ctest . --verbose -j -C ${{ matrix.build-type }}'
```
EduMenges, line 578: Should we keep this?

src/account/AccountMessenger.cpp
```diff
@@ -1123,36 +1308,154 @@ namespace sgns
...
+        {
+            if ( !has_utxos )
+            {
+                return "none";
+            }
+            std::string key;
+            for ( const auto &u : utxos )
+            {
+                key.append( u );
+                key.push_back( '\x1f' );
```
EduMenges: Why this specific character?

src/account/GeniusAccount.cpp
```diff
@@ -621,6 +688,29 @@ namespace sgns
...
+        }
+        genius_account_logger()->debug( "Fetching nonce from the network with timeout {} ms", timeout_ms );
+
+        auto result = messenger_->GetLatestNonce( timeout_ms );
+        if ( result.has_value() )
+        {
+            genius_account_logger()->debug( "Nonce replied with value {}", result.value() );
+            return result.value();
+        }
+        else if ( result.error() == AccountMessenger::Error::RESPONSE_WITHOUT_NONCE )
```
EduMenges:
```suggestion
        if ( result.error() == AccountMessenger::Error::RESPONSE_WITHOUT_NONCE )
```

src/account/GeniusAccount.cpp
```diff
@@ -790,4 +907,120 @@ namespace sgns
...
+
+            auto address = key.substr( std::string( NONCE_KEY_PREFIX ).size() );
+            try
+            {
+                uint64_t nonce  = std::stoull( std::string( val_buf.toString() ) );
+                loaded[address] = nonce;
+            }
+            catch ( const std::exception & )
+            {
+            }
```
EduMenges, line 994: Empty on purpose?

src/account/TransactionManager.cpp
```diff
@@ -1582,56 +1567,267 @@ namespace sgns
...
+        if ( missing_count == 0 )
+        {
+            if ( CheckNonce() )
             {
                 ChangeState( State::READY );
             }
+            return;
+        }
+        // TODO - Remove this once we remove the passive heads processing or we want transactions we are not subscribed here
+        return;
```
EduMenges, line 1727: Still valid?

EduMenges (changes requested):

.github/workflows/cmake.yml
henriqueaklein: Specially with debug it helped on finding issues. I don't see it as an issue

src/account/AccountMessenger.cpp
henriqueaklein: This is a separator only. I don't think it's needed

src/account/GeniusAccount.cpp
henriqueaklein: good catch

src/account/TransactionManager.cpp
henriqueaklein: This early return is to avoid the active fetching of the transactions. As I mentioned, we should remove this. I thought of encapsulating the code below in a method and removing but then I might forget to use it afterwards

src/account/AccountMessenger.cpp
henriqueaklein: The only reason it's needed is if the UTXOs aren't fixed length and not a hash, so it would be used to differentiate concatenation of different strings matching the same resulting string

EduMenges (approved):

PR #245, opened 2026-03-13
Removed unused files

EduMenges:


example/CMakeLists.txt
```diff
@@ -9,4 +9,3 @@ add_subdirectory(account_handling)
 add_subdirectory(node_test)
 add_subdirectory(mnn_chunkprocess)
 add_subdirectory(processing_json)
-add_subdirectory(evm_messaging_dapp)
```
henriqueaklein, line 12: We should keep this as it was kinda recent.

src/watcher/messaging_watcher.hpp
```diff
@@ -1,48 +0,0 @@
-#ifndef MESSAGING_WATCHER_HPP
```
henriqueaklein, line 1: This was also somewhat recent

src/CMakeLists.txt
```diff
@@ -8,11 +7,9 @@ add_subdirectory(primitives)
 add_subdirectory(crypto)
 add_subdirectory(storage)
 add_subdirectory(crdt)
-add_subdirectory(runtime)
 add_subdirectory(subscription)
 add_subdirectory(processing)
 add_subdirectory(local_secure_storage)
 add_subdirectory(singleton)
-add_subdirectory(watcher)
```
henriqueaklein, line 16: keep watcher

test/src/watcher/messaging_watcher_test.cpp
```diff
@@ -1,137 +0,0 @@
-#include <gtest/gtest.h>
```
henriqueaklein, line 1: keep the watcher test

henriqueaklein (changes requested):

Issue #251, opened 2026-03-20
Need to fix possible null pointer crash


Super-Genius:
### 💡 Codex Review

https://github.com/GeniusVentures/SuperGenius/blob/3106c206cf0fd16aef31ab3150004152aa8900b4/src/account/GeniusAccount.cpp#L186-L189
**<sub><sub>![P1 Badge](https://img.shields.io/badge/P1-orange?style=flat)</sub></sub>  Restore account creation in path-only New overload**

`GeniusAccount::New(token_id, base_path, full_node)` now returns `nullptr` whenever no stored key is found, instead of generating a new account as before. This breaks fresh-node startup paths: `GeniusNode::New(dev_config, ...)` still calls this overload and then the constructor dereferences `account_` (`account_->GetAddress()`), so a clean data directory can crash during node creation rather than initializing a new account.
    

<details> <summary>ℹ️ About Codex in GitHub</summary>
<br/>

[Your team has set up Codex to review pull requests in this repo](http://chatgpt.com/codex/settings/general). Reviews are triggered when you
- Open a pull request for review
- Mark a draft as ready
- Comment "@codex review".

If Codex has suggestions, it will comment; otherwise it will react with 👍.

Codex can also answer questions or update the PR. Try commenting "@codex address that feedback".
            
</details>

_Originally posted by @chatgpt-codex-connector[bot] in https://github.com/GeniusVentures/SuperGenius/issues/250#issuecomment-4093623440_

PR #252, opened 2026-03-20
Added account generation from random mnemonic if no previous address was

EduMenges:
Closes [#251]

### DIFF
--- a/src/account/GeniusNode.cpp
+++ b/src/account/GeniusNode.cpp
@@ -172,23 +172,19 @@ namespace sgns
         return instance;
     }
 
-    std::shared_ptr<GeniusNode> GeniusNode::NewWithMnemonics( const DevConfig_st &dev_config,
-                                                              const char         *mnemonics,
-                                                              bool                autodht,
-                                                              bool                isprocessor,
-                                                              uint16_t            base_port,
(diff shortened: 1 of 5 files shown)


src/account/GeniusNode.cpp
```diff
@@ -172,23 +172,19 @@ namespace sgns
...
         try
         {
-            TW::HDWallet wallet( mnemonics, "", true );
-            auto         derivation_path = TW::derivationPath( TWCoinTypeEthereum );
-            auto         private_key     = wallet.getKey( TWCoinTypeEthereum, derivation_path );
-
             auto instance = std::shared_ptr<GeniusNode>( new GeniusNode(
                 dev_config,
-                GeniusAccount::New( dev_config.TokenID, private_key, dev_config.BaseWritePath, is_full_node ),
+                GeniusAccount::NewFromMnemonic( dev_config.TokenID, mnemonic, dev_config.BaseWritePath, is_full_node ),
```
henriqueaklein: It doesn't check for nullptr. Probaly best to create the account first, check for nullptr and if not, create the GeniusNode instance

EduMenges: K

PR #259, opened 2026-04-16
Dev transfer account

EduMenges:


src/account/GeniusNode.cpp
```diff
@@ -1780,10 +1856,7 @@ namespace sgns
         switch ( new_state )
         {
             case TransactionManager::State::READY:
-                if ( isprocessor_ )
-                {
-                    StartProcessing();
-                }
+                StateTransition( NodeState::INITIALIZING_PROCESSING );
```
henriqueaklein: You should check if the processing is already initialized. The whole point of these states is so that the SYNCING makes it stop and the ready makes it continue. Right now if we lose a nonce and stop and restart it's going to try to recreate the processing.

src/account/GeniusNode.cpp
```diff
@@ -1014,6 +1030,65 @@ namespace sgns
...
+    outcome::result<void> GeniusNode::MergeAccount( std::string_view public_address )
+    {
+        BOOST_OUTCOME_TRY( TransferAccount( public_address ) );
+        return DeleteAccount( public_address );
+    }
+
+    outcome::result<void> GeniusNode::SetPayoutAddress( std::string_view payout_address )
+    {
+        BOOST_OUTCOME_TRY( account_->SaveInSecureStorage( "payout_address", std::string( payout_address ) ) );
+
```
henriqueaklein, line 1088: These ought to be a method, no? after all it's used in another place. Also evaluate if it isn't better to have on INITIALIZING_PROCESSING instead, so it would just make sure it's empty

test/src/transaction_sync/transaction_sync_test.cpp
```diff
@@ -55,28 +56,34 @@ namespace sgns
...
+                                               "9389e5f08c01e791dc436abab7a61a502515ddc7f91cb09f10289e147c651780",
                                                false,
                                                false,
                                                40001,
                                                true );
-            test::assertWaitForCondition(
-                [&]() { return full_node->GetTransactionManagerState() == TransactionManager::State::READY; },
-                std::chrono::milliseconds( 40000 ),
-                "full_node not ready" );
+            Blockchain::SetAuthorizedFullNodeAddress( full_node->GetAddress() );
```
henriqueaklein, line 75: Does the setting of the authorized node works after creating it?

src/account/GeniusNode.cpp
EduMenges: Good catch

src/account/GeniusNode.cpp
EduMenges: I have not understood what you said

test/src/transaction_sync/transaction_sync_test.cpp
EduMenges: Yes (on my machine it works)

src/account/GeniusNode.cpp
henriqueaklein: The cleanup of processing_service_ and whatnot could be inside a method and this method could be called on the start of INITIALIZING_PROCESSING, because right now you manually reset those 4 members before you go to INITIALIZING_PROCESSING. That way you guarantee they are reset when you create the processing_service_ again.

henriqueaklein (approved):

PR #299, opened 2026-06-01
fix(tests): use in-memory secure storage to eliminate keychain prompts

Super-Genius:
## Summary

Replaces macOS keychain-backed secure storage with in-memory storage for test builds. Eliminates ~20 keychain password prompts per test run.

## Changes

- New: `src/local_secure_storage/impl/MemorySecureStorage.hpp` — in-memory JSON backend, no OS keychain access
- Updated: `SecureStorage.hpp` — respects `SGNS_USE_MEMORY_SECURE_STORAGE` compile flag
- Updated: `CommonBuildParameters.cmake` — adds flag before library build when `BUILD_TESTING=ON`
- Updated: `secure_storage_test.cpp` — uses MemorySecureStorage on macOS

## How it works

When `BUILD_TESTING=ON`, the cmake build adds `-DSGNS_USE_MEMORY_SECURE_STORAGE` globally. This causes `SecureStorageImpl` to resolve to `MemorySecureStorage` (in-memory map) instead of `AppleSecureStorage` (macOS keychain). Both library and test targets get the flag.

## Testing

```bash
cd build/OSX/Debug
cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Debug
ninja -j4
ctest --output-on-failure -j4
```

Zero keychain prompts expected.

### DIFF
--- a/src/account/CMakeLists.txt
+++ b/src/account/CMakeLists.txt
@@ -53,8 +53,8 @@ if(MSVC)
 elseif(ANDROID_ABI STREQUAL "armeabi-v7a")
     target_compile_options(sgns_genius_account PUBLIC -fconstexpr-steps=10000000)
 endif()
-add_library(genius_node
-    IGeniusTransactions.cpp
+set(GENIUS_NODE_SOURCES
+    GeniusTransaction.cpp
     TransferTransaction.cpp
     MintTransaction.cpp
     MintTransactionV2.cpp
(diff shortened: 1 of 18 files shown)


Super-Genius:
@codex review

henriqueaklein:
This is duplicated. Already implemented in phase5 branch

PR #308, opened 2026-06-01
Bring crdt backup, config, and network config logic from 3.5 branch.

itsafuu:


src/crdt/globaldb/globaldb.cpp
```diff
@@ -122,12 +124,16 @@ namespace sgns::crdt
         std::shared_ptr<sgns::ipfs_lite::ipfs::graphsync::Network>            graphsyncnetwork,
         std::shared_ptr<libp2p::basic::Scheduler>                             scheduler,
         std::shared_ptr<sgns::ipfs_lite::ipfs::graphsync::RequestIdGenerator> generator,
-        std::shared_ptr<RocksDB>                                              datastore )
+        std::shared_ptr<RocksDB>                                              datastore,
+        BackupOptions                                                         backup_options )
     {
+        backup_options_ = backup_options;
```
henriqueaklein: It would be better if backup_options_ initialization was on the constructor.

src/crdt/globaldb/globaldb.cpp
```diff
@@ -155,41 +161,83 @@ namespace sgns::crdt
...
 
-                        if ( dataStoreResult.has_value() )
+                        if ( !restoreFromLatestBackup() )
                         {
-                            break;
+                            m_logger->critical(
+                                "Restore from backup failed for corrupted DB {}. "
+                                "Terminating process to preserve on-disk state for forensic analysis.",
+                                databasePathAbsolute );
+                            std::abort();
```
henriqueaklein, line 238: This was by design, right? The abort

src/crdt/globaldb/globaldb.hpp
```diff
@@ -218,6 +239,13 @@ namespace sgns::crdt
         bool                                              cid_sync_started_;
         bool                                              cid_receiving_started_;
         bool                                              head_broadcasting_started_;
+        BackupOptions                                     backup_options_{};
+        std::string                                       backup_directory_;
+        std::atomic_bool                                  stop_backup_thread_{ false };
+        std::thread                                       backup_thread_;
+        std::mutex                                        backup_mutex_;
```
henriqueaklein: is backup_mutex_ necessary? It's only used in one place and it seems no more than a single thread executes the code that uses it

src/crdt/globaldb/globaldb.hpp
```diff
@@ -229,30 +257,14 @@ namespace sgns::crdt
...
-            //"/dnsaddr/fra1-2.hostnodes.pinata.cloud/ipfs/QmNfpLrQQZr5Ns9FAJKpyzgnDL2GgC6xBug1yUZozKFgu4",
-            //"/dnsaddr/fra1-3.hostnodes.pinata.cloud/ipfs/QmPo1ygpngghu5it8u4Mr3ym6SEU2Wp2wA66Z91Y1S1g29",
-            //"/dnsaddr/nyc1-1.hostnodes.pinata.cloud/ipfs/QmRjLSisUCHVpFa5ELVvX3qVPfdxajxWJEHs9kN3EcxAW6",
-            //"/dnsaddr/nyc1-2.hostnodes.pinata.cloud/ipfs/QmPySsdmbczdZYBpbi2oq2WMJ8ErbfxtkG8Mo192UHkfGP",
-            //"/dnsaddr/nyc1-3.hostnodes.pinata.cloud/ipfs/QmSarArpxemsPESa6FNkmuu9iSE1QWqPX2R3Aw6f5jq4D5",
-        };
+        /** @brief Resolves the backup directory path based on the database path. */
+        std::string ResolveBackupDirectory( const std::string &databasePathAbsolute ) const;
+        /** @brief Creates a backup immediately. */
+        bool        CreateBackupNow();
```
henriqueaklein: no need for boolean return on CreateBackupNow if not used. Consider changing it to void or using the return value.

src/account/GeniusNode.hpp
```diff
@@ -744,6 +751,16 @@ namespace sgns
         std::atomic<NodeState> state_{ NodeState::CREATING }; ///< Current node lifecycle state.
         bool                   use_upnp_;                     ///< Whether UPnP mapping is enabled.
 
+        struct CrdtBackupConfig
```
henriqueaklein: It's the same struct from globaldb, right? Can't we use that definition instead of duplicating it?

src/account/GeniusNode.cpp
```diff
@@ -262,19 +264,68 @@ namespace sgns
         }
         node_logger_->debug( "Account Address {}", account_->GetAddress() );
 
-        // Initializes the thread pool for IO context
-        io_threads_.reserve( io_thread_count_ );
-        for ( unsigned i = 0; i < io_thread_count_; ++i )
+        LoadCrdtConfig();
+    }
```
henriqueaklein, line 267: io_threads_ initialization lost in the merge

src/account/GeniusNode.hpp
```diff
@@ -607,6 +607,8 @@ namespace sgns
         std::string                 gnus_network_full_path_;       ///< Versioned network DB path.
         std::string                 processing_channel_topic_;     ///< Processing task channel topic.
         std::string                 processing_grid_chanel_topic_; ///< Processing grid topic.
+        std::vector<std::string>                              bootstrap_peers_;
```
henriqueaklein, line 611: bootstrap_peers_ doesn't need to be a member as it's used only on the InitNetwork, it could be just a local variable there.

src/account/GeniusNode.cpp
```diff
@@ -608,73 +664,136 @@ namespace sgns
...
     bool GeniusNode::InitNetwork( uint16_t base_port, bool is_full_node )
     {
-        bool ret    = true;
-        pubsubport_ = GenerateRandomPort( base_port, account_->GetAddress() );
+        bool ret = true;
+        std::string config_path = write_base_path_ + "/network_config.json";
+        rapidjson::Document config_json;
+        std::string pubsub_bind_address = "0.0.0.0";
+        std::string authorized_full_node;
+        bool upnp_enabled = use_upnp_;
```
henriqueaklein: upnp_enabled is not really necessary, is it?

henriqueaklein (changes requested):
io_threads_ is probably the only error. The rest are suggestions/improvements.

src/crdt/globaldb/globaldb.cpp
itsafuu: Yes, the repair path will actually destroy parts of the database if unrecoverable. If we want to try to find a root cause I would like to get a snapshot of the database from the tester.

src/account/GeniusNode.cpp
itsafuu: I would say that use_upnp_ is not necessary anymore.

src/account/GeniusNode.hpp
itsafuu: Originally full node was part of bootstrap peers, however an upcoming update has reconnect logic for the full node, and i'll probably also include the bootstrap nodes. The purpose of the reconnect logic is for mobile devices that might drop connections while app is in the background. That logic is mostly done as is, so i'll probably add that to this.

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+
+        std::vector<libp2p::multi::Multiaddress> multiaddresses;
+        multiaddresses.push_back( std::move( ma ) );
+
+        return libp2p::peer::PeerInfo{ peer_id_res.value(), std::move( multiaddresses ) };
+    }
+
+    void GeniusNode::InitBootstrapReconnect()
+    {
+        if ( bootstrap_fullnode_ids_.empty() )
```
henriqueaklein: if ( bootstrap_fullnode_ids_.empty() ) seems duplicated

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+                    [weak_self = weak_from_this()]( const libp2p::peer::PeerId &peer_id )
+                    {
+                        if ( auto strong = weak_self.lock() )
+                        {
+                            if ( strong->shutdown_started_.load() )
+                            {
+                                return;
+                            }
+                            bool is_bootstrap = false;
+                            if ( strong->bootstrap_fullnode_ids_.count( peer_id ) )
```
henriqueaklein: if ( strong->bootstrap_fullnode_ids_.count( peer_id ) ) seems a bit weird, since you use it on the is_bootstrap also. Is this correct?

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+                                    {
+                                        attempt = it->second;
+                                    }
+                                }
+                                strong->ScheduleBootstrapReconnect( peer_id, attempt );
+                            }
+                        }
+                    } ) );
+
+        node_logger_->info( "Subscribed to disconnect events for {} bootstrap fullnode(s)",
```
henriqueaklein: Duplicated log?

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+        return libp2p::peer::PeerInfo{ peer_id_res.value(), std::move( multiaddresses ) };
+    }
+
+    void GeniusNode::InitBootstrapReconnect()
+    {
+        if ( bootstrap_fullnode_ids_.empty() )
+        if ( bootstrap_fullnode_ids_.empty() && bootstrap_peer_ids_.empty() )
+        {
+            node_logger_->debug( "No bootstrap fullnodes configured, skipping reconnect subscription" );
+            node_logger_->debug( "No bootstrap peers configured, skipping reconnect subscription" );
```
henriqueaklein, line 2277: Duplicated log?

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+                            bootstrap_peer_ids_.size() );
+    }
+
+    void GeniusNode::StartBootstrapHealthCheck()
+    {
+        if ( bootstrap_fullnode_infos_.empty() )
+        if ( bootstrap_fullnode_infos_.empty() && bootstrap_peer_infos_.empty() )
+        {
+            node_logger_->debug( "No bootstrap fullnodes to health-check" );
+            node_logger_->debug( "No bootstrap peers to health-check" );
```
henriqueaklein, line 2328: Duplicated log?

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+            return;
+        }
+        ScheduleNextHealthCheck();
+        node_logger_->info( "Bootstrap fullnode health check started (interval: {}s)",
```
henriqueaklein: Duplicated log?

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+        node_logger_->info( "Subscribed to disconnect events for {} bootstrap fullnode(s)",
+                            bootstrap_fullnode_ids_.size() );
+        node_logger_->info( "Subscribed to disconnect events for {} bootstrap fullnode(s) + {} peer(s)",
+                            bootstrap_fullnode_ids_.size(),
+                            bootstrap_peer_ids_.size() );
+    }
+
+    void GeniusNode::StartBootstrapHealthCheck()
+    {
+        if ( bootstrap_fullnode_infos_.empty() )
```
henriqueaklein: extra if ( bootstrap_fullnode_infos_.empty() )

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+        {
+            std::lock_guard<std::mutex> lock( reconnect_mutex_ );
+            if ( !reconnect_attempts_.empty() )
+            {
+                interval = reconnect_config_.health_check_disconnected_interval;
+            }
+        }
+
+        // Apply background multiplier
+        if ( background_mode_.load() )
```
henriqueaklein: background_mode_ is never set, is that correct?

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+
+    void GeniusNode::PerformHealthCheck()
+    {
+        if ( shutdown_started_.load() )
+        {
+            return;
+        }
+
+        auto host = pubsub_->GetHost();
+        for ( const auto &peer_info : bootstrap_fullnode_infos_ )
```
henriqueaklein: This seems extra.

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+
+        // Check both fullnodes and peers
+        for ( const auto &infos : { &bootstrap_fullnode_infos_, &bootstrap_peer_infos_ } )
+        {
+            auto connectedness = host->connectedness( peer_info );
```
henriqueaklein: This connectedness stuff seems duplicated

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+            if ( connectedness == libp2p::Host::Connectedness::NOT_CONNECTED ||
+                 connectedness == libp2p::Host::Connectedness::CAN_NOT_CONNECT )
+            for ( const auto &peer_info : *infos )
+            {
+                node_logger_->debug( "Health check: bootstrap fullnode {} is {}",
```
henriqueaklein: This log seems duplicated

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+                        std::lock_guard<std::mutex> lock( reconnect_mutex_ );
+                        auto                          it = reconnect_attempts_.find( peer_info.id );
+                        if ( it != reconnect_attempts_.end() )
+                        {
+                            attempt = it->second;
+                        }
+                    }
+                    ScheduleBootstrapReconnect( peer_info.id, attempt );
+                }
+                ScheduleBootstrapReconnect( peer_info.id, attempt );
```
henriqueaklein: This seems duplicated

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+    {
+        if ( shutdown_started_.load() )
+        {
+            return;
+        }
+
+        // Find the PeerInfo for this peer_id
+        // Find the PeerInfo for this peer_id (search fullnodes then peers)
+        const libp2p::peer::PeerInfo *peer_info_ptr = nullptr;
+        for ( const auto &info : bootstrap_fullnode_infos_ )
```
henriqueaklein: This is duplicated

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+        for ( const auto &infos : { &bootstrap_fullnode_infos_, &bootstrap_peer_infos_ } )
+        {
+            if ( info.id == peer_id )
```
henriqueaklein: This is duplicated

src/account/GeniusNode.cpp
```diff
@@ -1950,4 +2185,333 @@ namespace sgns
...
+            {
+                if ( info.id == peer_id )
+                {
+                    peer_info_ptr = &info;
+                    break;
+                }
+            }
+            if ( peer_info_ptr )
+            {
+                peer_info_ptr = &info;
```
henriqueaklein: This seems to be duplicated

src/crdt/impl/crdt_datastore.cpp
```diff
@@ -465,29 +465,88 @@ namespace sgns::crdt
...
         bool expected = false;
-        if ( !closeStarted_.compare_exchange_strong( expected, true ) )
+        if ( !shutdown_started_.compare_exchange_strong( expected, true ) )
         {
+            logger_->warn( "CancelAndCloseNow called but shutdown has already started" );
             return;
         }
 
-        StopWorkerLoops();
+        {
```
henriqueaklein: IsCurrentThreadInternalWorker needs to be called and StopWorkerLoops too. This happens when a worker thread is the last reference to a crdt_datastore. When that happens it tries to delete itself by notifying itself and gets stuck

henriqueaklein (changes requested):
Check the duplicated code (probably result of merge errors) and more importantly, the crdt_datastore destructor needs the code that detects if the destructor was called by a worker thread, otherwise it will get stuck

src/account/GeniusNode.cpp
itsafuu: I think this may have been for android background mode, but it's not actually wired to anything, removing until later.

henriqueaklein (approved):

PR #315, opened 2026-06-16
fix(05): chain RPC fixes + BridgeSourceBurned 6-param update

Super-Genius:
### Chain RPC / bridge config fixes
- **Decouple chainlist_json_path from bridge_chains_path** — bridge_chains_config.json contains bridge contract addresses only, not a ChainList JSON array. Setting chainlist_json_path to the bridge config path caused zero RPC endpoints to be wired.
- **Disambiguate naming** — rename chains_json_path → chainlist_json_path, config file → bridge_chains_config.json. Add bundled default config.
- **Catch-up scan from_block fix** — query eth_blockNumber before eth_getLogs to compute a relative from_block instead of using scan_depth as an absolute block number.
- **Honor ChainRpcEndpointProvider::Initialize return value** — check the outcome instead of silently swallowing failure.

### Bridge event signature update
- **Update BridgeSourceBurned from 5 params → 6 params** to match new [GNUSBridge.sol](https://github.com/GeniusVentures/TokenContracts/blob/develop/gnus-ai/contracts/gnus-ai/GNUSBridge.sol) which adds `bytes sgnsDestination` (64-byte SuperGenius public key).
- BridgeRelayer now decodes values[5] as `codec::ByteBuffer`, converts via `hex_bytes`, and uses it as the mint destination instead of the EVM sender address.
- Updated event_registry in evmrelay submodule (pushed to evmrelay develop).
- Updated all test references to the new 6-param signature.

### Files changed
- `src/account/BridgeRelayer.cpp` — decode 6th param, remove AddressToHex
- `src/account/GeniusNode.cpp` — event_sig strings, from_block fix, Initialize return check
- `test/src/account/bridge_relayer_test.cpp` — MakeBurnNotification 6-param, new round-trip test
- `test/src/startup/startup_wiring_test.cpp` — event sig constant
- `test/src/account/chain_rpc_endpoint_provider_test.cpp` — event sig strings
- `test/src/watcher/messaging_watcher_test.cpp` — event sig strings
- `evmrelay/include/eth/eth_watch_cli.hpp` — registered 6th param (kBytes)

### DIFF
--- a/src/account/BridgeRelayer.cpp
+++ b/src/account/BridgeRelayer.cpp
@@ -30,16 +30,6 @@ namespace sgns
     namespace
     {
         /**
-         * @brief       Auxiliary function to convert an eth::Address to a hex string for logging.
-         * @param[in]   addr The eth::Address to convert.
-         * @return      Address as a hex string.
-         */
-        std::string AddressToHex( const eth::Address &addr )
-        {
-            return rlp::base::parse::hex_array_string( addr );
-        }
-
-        /**
          * @brief       Converts an intx::uint256 to uint64_t with overflow check, used for event parameters that must fit in uint64.
          * @param[in]   value The intx::uint256 value to convert.
          * @param[in]   field The name of the field being converted.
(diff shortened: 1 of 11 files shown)


Super-Genius:
## Status: 🔖 Ready (partially blocked)

**Chain RPC fixes (commits d989749a–db3599e6):** Ready for review and merge.

**6-param BridgeSourceBurned update (commit 81e8a7ae):** Code is complete but **live E2E testing is blocked** until the new GNUSBridge.sol is deployed with the `bytes sgnsDestination` field. The PR can land — the event signature change is gated by the contract deployment, and until then no 6-param events will be emitted on-chain.

Super-Genius:
Superseded by #314 — all 5 commits on this branch are already included in `explore_fix_chain_rpc_json` (merged at 53ebe0b9) and reach `bridge_phase5` via #314. Consolidating into a single reviewable PR per the two-stage plan: #314 → `bridge_phase5` (this review), then `bridge_phase5` → `develop` as a separate reviewed PR.

PR #321, opened 2026-06-22
Align inits with GeniusWallet

EduMenges:


src/account/GeniusNode.cpp
```diff
@@ -183,33 +181,52 @@ namespace sgns
...
 
-            if ( instance )
-            {
-                instance->BeginDBInitialization();
-            }
+        return instance;
+    }
 
-            return instance;
+    std::pair<std::shared_ptr<GeniusNode>, std::string> GeniusNode::NewFromRandomMnemonic(
```
henriqueaklein: Not sure if necessary

src/account/GeniusNode.hpp
```diff
@@ -118,6 +118,22 @@ namespace sgns
...
+        /**
+         * @brief Creates a node with a newly generated random BIP39 mnemonic.
+         * @param[in] dev_config Runtime configuration for paths, token settings, and payout data.
+         * @param[in] autodht Whether to start DHT discovery.
+         * @param[in] isprocessor Whether this node should run processing services.
+         * @param[in] base_port Base pubsub port used to derive the node listening port.
+         * @param[in] is_full_node Whether the node should run in full-node mode.
+         * @return Pair of shared node instance (nullptr on failure) and the generated mnemonic phrase.
+         */
+        static std::pair<std::shared_ptr<GeniusNode>, std::string> NewFromRandomMnemonic(
```
henriqueaklein: Not necessary I think

henriqueaklein (changes requested):
Evaluate if we need to externalize a new from random mnemonic on the genius Node

henriqueaklein (approved):

PR #324, opened 2026-06-26
Refactor transaction manager

EduMenges:


src/account/TransactionManager.cpp
```diff
@@ -128,12 +146,9 @@ namespace sgns
                     auto process_result = strong->OnConsensusCertificate( subject_hash );
                     if ( process_result.has_error() )
                     {
-                        TransactionManagerLogger()->error(
-                            "[{} - full: {}] Failed to process certificate proposal_id={} error={}",
-                            strong->account_m->GetAddress().substr( 0, 8 ),
-                            strong->full_node_m,
-                            certificate.proposal_id(),
-                            process_result.error().message() );
+                        strong->m_logger->error( "Failed to process certificate proposal_id={} error={}",
```
henriqueaklein, line 147: m_logger is problematic because it can be initialized before GeniusNode's TransactionManager logger, so it writes to console instead of file, for instance

src/account/TransactionManager.cpp
```diff
@@ -691,79 +636,85 @@ namespace sgns
...
         BOOST_OUTCOME_TRY( auto peers_cut_ptr, TokenAmount::New( escrow_tx->GetPeersCut() ) );
 
         BOOST_OUTCOME_TRY( auto peer_total, escrow_amount_ptr->Multiply( *peers_cut_ptr ) );
 
-        const auto escrowTokenId = escrow_tx->GetUTXOParameters().second[0].token_id;
+        const auto subtask_count   = static_cast<uint64_t>( subtask_results.size() );
+        const auto peers_amount    = peer_total.Value() / subtask_count;
+        const auto peer_total_paid = peers_amount * subtask_count;
+        const auto escrow_token_id = escrow_params.second.front().token_id;
+        if ( peer_total_paid > escrow_amount )
```
henriqueaklein, line 796: can't this run into a warning for mismatch types?

src/account/TransactionManager.cpp
```diff
@@ -812,259 +761,169 @@ namespace sgns
...
-
-        auto tracked_hash = GetTrackedOutgoingPreviousHash( nonce );
-        if ( !tracked_hash.empty() )
-        {
-            return tracked_hash;
-        }
-
-        auto persisted_hash = GetPersistedOutgoingPreviousHash( nonce );
-        if ( !persisted_hash.empty() )
+        const auto               previous_hash = [&]() -> std::string
```
henriqueaklein, line 878: that lambda is more confusing then the original 2 separate methods

henriqueaklein (approved):

PR #328, opened 2026-07-03
Fix init parameters

henriqueaklein:
- Removed parameters for the creation of the node. Migrated all of them to wither network_config.json or sgns_config.json.
- Reduced the creation methods to just a single one with a variant and the dev_config, to make it simpler for the SDK and therefore the final app


src/account/GeniusAccount.cpp
```diff
@@ -1268,13 +1268,22 @@ namespace sgns
 
     void GeniusAccount::SetNonceStore( std::shared_ptr<storage::rocksdb> db )
     {
-        nonce_db_ = std::move( db );
+        {
+            std::lock_guard lock( nonce_db_mutex_ );
+            nonce_db_ = std::move( db );
+        }
```
EduMenges, line 1274: Why tho?

src/account/GeniusNode.hpp
```diff
@@ -70,61 +71,62 @@ namespace sgns
...
+         * @brief Writes a minimal network_config.json for test/example setup (MIG-02).
+         * @param[in] base_path Directory whose network_config.json will be (over)written (dev_config.BaseWritePath).
+         * @param[in] port_seed Numeric port seed (Phase-1 key "port_seed").
+         * @param[in] auto_dht  Whether DHT discovery is enabled (key "auto_dht").
+         * @return Failure on file I/O error; success otherwise. Truncates/rewrites the file (deterministic minimal config).
+         */
+        static outcome::result<void> WriteNetworkConfig( const std::string &base_path,
+                                                        uint16_t            port_seed,
+                                                        bool                auto_dht );
+
```
EduMenges, line 132: Does it make sense for this to be here? Looks more like a test helper

src/account/GeniusNode.hpp
```diff
@@ -679,7 +723,8 @@ namespace sgns
         std::shared_ptr<soralog::LoggingSystem>               logging_system_;      ///< libp2p logging system.
         bool                                                  autodht_;     ///< Whether DHT discovery is enabled.
         bool                                                  isprocessor_; ///< Whether processing service should run.
-        bool                           is_full_node_;              ///< Whether this node runs in full-node mode.
+        bool                                     is_full_node_ = false;   ///< Whether this node runs in full-node mode.
+        NodeType                                 node_type_ = NodeType::Light; ///< Role from sgns_config.json (default Light; derived in the AccountSource ctor).
```
EduMenges: Incorrect formatting

src/account/GeniusNode.cpp
```diff
@@ -94,6 +96,25 @@ namespace
...
+
+    // Case-insensitive parse of the "node_type" sgns_config.json value (CONTEXT D-02).
+    // Returns nullopt for unrecognized values; the caller (LoadSgnsConfig) WARN-logs + defaults to Light.
+    std::optional<sgns::GeniusNode::NodeType> NodeTypeFromString( std::string_view s )
+    {
+        std::string lower;
+        lower.reserve( s.size() );
+        for ( char c : s )
+        {
+            lower.push_back( static_cast<char>( std::tolower( static_cast<unsigned char>( c ) ) ) );
```
EduMenges, line 108: Why the weird casting?

src/account/GeniusNode.cpp
```diff
@@ -143,92 +166,70 @@ namespace sgns
...
-    GeniusNode::GeniusNode( const DevConfig_st            &dev_config,
-                            std::shared_ptr<GeniusAccount> account,
-                            bool                           autodht,
-                            uint16_t                       base_port,
-                            bool                           is_full_node ) :
+    // Reordered constructor (INTF-03 / CONTEXT D-05). Account is created via std::visit
+    // AFTER LoadSgnsConfig() resolves node_type_ -> is_full_node_ (the init-order hinge fix).
+    // account_ and is_full_node_ are default-init here (no source/param) and assigned in the
+    // body; autodht_ defaults to true (Phase-1 config layer overrides from network_config.json).
+    // Throws on account-restore failure; New(dev_config, AccountSource) catches -> nullptr (D-04).
```
EduMenges, line 226: These comments are obnoxious

src/account/GeniusNode.cpp
```diff
@@ -143,92 +166,70 @@ namespace sgns
...
+    GeniusNode::GeniusNode( const DevConfig_st &dev_config, AccountSource source ) :
         write_base_path_( dev_config.BaseWritePath ),
-        account_( std::move( account ) ),
         io_( std::make_shared<boost::asio::io_context>() ),
         io_work_guard_( boost::asio::make_work_guard( *io_ ) ),
-        autodht_( autodht ),
+        autodht_( true ),
```
EduMenges, line 231: Warum?

src/account/GeniusNode.cpp
```diff
@@ -250,9 +251,49 @@ namespace sgns
...
+                                                           is_full_node_ );
+                }
+                else if constexpr ( std::is_same_v<T, FromPublicKey> )
+                {
+                    // FromPublicKey carries a public_address; GeniusAccount::NewFromPublicKey
+                    // takes no base_path and consumes an address-like string_view.
+                    return GeniusAccount::NewFromPublicKey( dev_config_.TokenID,
+                                                            src.public_address,
+                                                            is_full_node_ );
+                }
```
EduMenges, line 287: Why didn't you also refactor `GeniusAccount` to have only one constructor?

AGENTS.md
```diff
@@ -0,0 +1,417 @@
...
+
+<!-- GSD:stack-start source:codebase/STACK.md -->
+## Technology Stack
+
+## Languages
+- C++ (C++17) - Entire core project: blockchain, consensus, processing, networking, crypto, storage
+- Python 3 - Test data generation scripts in `test/src/processing_datatypes/` (create models, convert NIfTI)
+- Java - Android keystore support in `src/local_secure_storage/impl/KeyStoreHelper.java`
+- JavaScript - Static documentation search in `docs/hdoc/`
+- Go - Used indirectly by `gRPCForSuperGenius` for gnostic/OpenAPI-to-protobuf toolchain
```
EduMenges: unused

AGENTS.md
```diff
@@ -0,0 +1,417 @@
...
+## Runtime
+- C++17 compiler required (GCC, Clang, or MSVC)
```
EduMenges: not GCC

AGENTS.md
```diff
@@ -0,0 +1,417 @@
...
+- No managed runtime — compiled to native binaries/libraries
+- Cross-compilation via CMake toolchains for Android (NDK r27b) and iOS
+- Git submodules (preferred dependency model — `thirdparty` repo at `../thirdparty/`)
+- CMake `find_package` with pre-built thirdparty libraries
+- Third-party dependency versions managed in `build/CommonBuildParameters.cmake`
+- npm (for QuickType code generation in `SGProcessingManager`)
```
EduMenges: ??

AGENTS.md
```diff
@@ -0,0 +1,417 @@
...
+- Vulkan (via MoltenVK on macOS) - GPU compute for processing tasks
+- zkLLVM - Zero-knowledge proof compiler infrastructure
+- Google Test (GTest) + Google Mock (GMock) - Unit testing framework
+- CMake 3.22+ - Build system generator
+- Ninja - Build executor (Linux/macOS)
+- Visual Studio 17 2022 - Build executor (Windows)
+- clang-format / clang-tidy - Code formatting and static analysis (`.clang-format`, `.clang-tidy`, `.clangd` present)
+## Key Dependencies
+| Package | Version | Purpose |
+|---------|---------|---------|
```
EduMenges: Isn't this repeating what is already in thirdparty?

EduMenges (changes requested):

src/account/GeniusAccount.cpp
henriqueaklein: because due to async calls, you can write on nonce_db_ a value that is already in use by one of the methods of the class in an async all, causing a nullptr dereference. The mutex is done so we can use it on the method to create a shared reference to it safely.

src/account/GeniusNode.hpp
henriqueaklein: it could a a test helper sure, but I left it here in case we wanted to use on GeniusSDK instead.

src/account/GeniusNode.cpp
henriqueaklein: In case of a nonce ASCII char.

src/account/GeniusNode.cpp
henriqueaklein: blame GSD

src/account/GeniusNode.cpp
henriqueaklein: It's the default value. if the configuration is not in the network_config file it defaults to true.

src/account/GeniusNode.cpp
henriqueaklein: This could be done, yes

AGENTS.md
henriqueaklein: you're right, I'm gonna remove this file entirely

src/account/GeniusAccount.cpp
EduMenges: K

src/account/GeniusNode.cpp
EduMenges: Kinda cringe but it's C++ I guess

src/account/GeniusNode.hpp
EduMenges: K

src/account/GeniusNode.cpp
EduMenges: I guess this could be done in a follow-up PR, since this one is already pretty big

EduMenges (approved):

PR #330, opened 2026-07-07
Asan tests

EduMenges:


src/blockchain/ValidatorRegistry.cpp
```diff
@@ -808,9 +807,14 @@ namespace sgns
             return outcome::success();
         }
 
-        if ( GetRegistryCid() != base_registry_cid || GetRegistryEpoch() != base_registry_epoch )
+        auto base_registry_result = LoadRegistryByCid( base_registry_cid );
+        if ( base_registry_result.has_error() )
+        {
```
henriqueaklein, line 812: This changes how the batchproposal is created. The previous version basically didn't allow for creation if the latest CID doesn't match it. Need to think if that's OK

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -879,14 +883,13 @@ namespace sgns
...
-        auto registry_result = LoadRegistry();
+        auto registry_result = LoadRegistryByCid( payload.base_registry_cid() );
         if ( registry_result.has_error() )
         {
             return outcome::success( BatchSubjectDecision::Pending );
         }
 
-        if ( registry_result.value().epoch() != payload.base_registry_epoch() ||
-             GetRegistryCid() != payload.base_registry_cid() )
+        if ( registry_result.value().epoch() != payload.base_registry_epoch() )
```
henriqueaklein, line 892: Again, this changes the evaluation to any past CID and not the current one. This might be better but need to think about it

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -905,42 +908,42 @@ namespace sgns
             {
                 return BatchCertificateDecision::Approve;
             }
+            applying_batch_subject_ids_.insert( subject_hash );
```
henriqueaklein, line 911: This generates confusion. This was done so that the erase afterwards always succeed, right?

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -1089,14 +1092,16 @@ namespace sgns
             return std::vector<crdt::pb::Element>{};
         }
 
-        RegistryUpdate  update      = decoded_update.value();
-        const Registry *current_ptr = nullptr;
+        RegistryUpdate          update = decoded_update.value();
+        std::optional<Registry> current_registry;
```
henriqueaklein: seems unfinished.

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -1172,30 +1195,54 @@ namespace sgns
...
 
-        if ( !current_registry )
+        const std::string prev_registry_cid = update.prev_registry_hash();
+        std::string       current_id;
+        {
+            std::shared_lock<std::shared_mutex> lock( cache_mutex_ );
+            current_id = cached_registry_id_;
+        }
+
+        std::optional<Registry> loaded_base_registry;
```
henriqueaklein: This optional seems unnecessary

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -1172,30 +1195,54 @@ namespace sgns
...
+        const Registry         *base_registry = current_registry;
```
henriqueaklein: not sure base_registry is needed

src/blockchain/ValidatorRegistry.cpp
```diff
@@ -1172,30 +1195,54 @@ namespace sgns
...
+        if ( !prev_registry_cid.empty() && ( !base_registry || current_id.empty() || prev_registry_cid != current_id ) )
```
henriqueaklein: This condition is a bit messy

src/blockchain/Consensus.cpp
```diff
@@ -644,16 +646,17 @@ namespace sgns
                 }
             }
 
-            if ( slot_state.best_proposal_id == proposal.proposal_id() && !slot_state.voted )
+            if ( slot_state.best_proposal_id == proposal.proposal_id() &&
```
henriqueaklein, line 649: This makes it so that I could vote more than once per subject_id, right? You need to explore that logic on the rest of the consensus algorithm.

henriqueaklein (changes requested):

src/blockchain/ValidatorRegistry.cpp
EduMenges: Yeah, this uses the batch CID registry instead of the node's registry. This can speed up things in case the node is behind.

src/blockchain/ValidatorRegistry.cpp
EduMenges: Pretty much

src/blockchain/ValidatorRegistry.cpp
EduMenges: Yep, very cringe

src/blockchain/Consensus.cpp
EduMenges: Yes

src/blockchain/ValidatorRegistry.cpp
EduMenges: 🃏

src/blockchain/ValidatorRegistry.cpp
EduMenges: Well the AI did it for deadlock preventions, but then again, AI is dumb sometimes

henriqueaklein (approved):

henriqueaklein (approved):

PR #340, opened 2026-07-14
Test and initialization speed up

EduMenges:


test/src/processing_nodes/processing_nodes_test.cpp
```diff
@@ -90,6 +85,11 @@ class ProcessingNodesTest : public ::testing::Test
 
         bootstrappers = { node_proc2->GetPubSub()->GetInterfaceAddress() };
         node_proc1->GetPubSub()->AddPeers( bootstrappers );
+
+        sgns::test::assertWaitForCondition( [&]
```
henriqueaklein, line 89: The whole point of waiting for it to be ready before was that the full node needs to answer the genesis for the other nodes, but if it works like this it's fine. Just gonna test a bit first

test/src/node/node_initialization_progress.cpp
```diff
@@ -36,13 +36,14 @@ TEST( GeniusNode, InitializationProgress )
...
-    ASSERT_WAIT_FOR_CONDITION(
-        [&]() { return node->GetState() == GeniusNode::NodeState::READY; },
-        std::chrono::milliseconds( 60000 ),
-        "Node did not reach READY state within timeout",
-        &elapsed );
+    auto end = std::chrono::steady_clock::now() + std::chrono::seconds(50);
+    while ( std::chrono::steady_clock::now() < end && node->GetState() != GeniusNode::NodeState::READY )
+    {
+        auto percentage = node->GetInitializationStatus().first;
+        ASSERT_GE( percentage, last_percentage );
```
henriqueaklein: sleep_for

henriqueaklein (changes requested):

test/src/processing_nodes/processing_nodes_test.cpp
EduMenges: The problem with waiting for the full node to be ready is that it piles up, so I moved its check later so that the other nodes can start their initialization after. Maybe this is problematic, but all the tests passed, so who knows.

test/src/node/node_initialization_progress.cpp
EduMenges: Yes, quite cringe, will change it on Monday 🤠

EduMenges: It turns out that it is necessary to use this pattern, because `assertWaitForCondition` does not allow for asserts inside the lambda's body

PR #346, opened 2026-07-17
Phase 08 - Create Burn/Mint race tests

henriqueaklein:


src/blockchain/ValidatorRegistry.hpp
```diff
@@ -408,6 +408,23 @@ namespace sgns
...
+         * (the common case for a Light node in a large concurrent cluster), it returns
+         * without requesting anything further and initialization is left to depend solely
+         * on a passive CRDT broadcast (RegistryUpdateReceived) that may never arrive for
+         * every node. Callers that periodically retry a deferred blockchain start (e.g.
+         * Blockchain::Start()) should call this on every such retry so an active,
+         * repeating head-CID request backs up the passive broadcast path.
+         *
+         * No-op once the cache is already initialized.
+         */
+        void RetryInitializationIfNeeded();
```
EduMenges, line 426: Where is this method called?

test/src/bridge_e2e/anvil_fixture.hpp
```diff
@@ -714,6 +743,44 @@ namespace sgns::test::anvil
...
+                }
+                acceptor.bind( { boost::asio::ip::tcp::v4(), static_cast<unsigned short>( candidate ) }, ec );
+                if ( !ec )
+                {
+                    acceptor.close( ec );
+                    return candidate;
+                }
+            }
+            return 0u;
+        }
```
EduMenges, line 782: Kinda cringe

test/src/bridge_race/bridge_race_fixture.hpp
```diff
@@ -0,0 +1,396 @@
...
+     *        measured runs exceed it).
+     */
+    static inline constexpr std::chrono::milliseconds kRaceNodeReadyTimeout{ 90000 };
+
+    /**
+     * @brief Stability window a test waits after observing the expected mint, to catch
+     *        a delayed double-mint on the watcher's next poll cycle (> production poll
+     *        interval).
+     */
+    static inline constexpr std::chrono::milliseconds kRaceStabilityWindow{ 16000 };
```
EduMenges: Do they need to be static inline? Also a `constexpr std::string_view` would be better than a raw char pointer

EduMenges (changes requested):

src/blockchain/ValidatorRegistry.hpp
henriqueaklein: on Blockchain::Start

test/src/bridge_e2e/anvil_fixture.hpp
henriqueaklein: Indeed, but it's used only for mocking. This was needed because an Anvil instance was running on another test causing the burn race test to grab the inputs created there. This allows another Anvil instance to be created in another port and avoid conflict with that one

test/src/bridge_race/bridge_race_fixture.hpp
henriqueaklein: "You're absolutely right!"

src/blockchain/ValidatorRegistry.hpp
EduMenges: K

src/account/GeniusNode.cpp
```diff
@@ -2965,16 +2970,18 @@ namespace sgns
         chainlist_fetcher_ = std::move( fetcher );
     }
 
-    void GeniusNode::ConfigureRpcEndpoint( const std::string &chain_id, std::vector<WeightedRpcEndpoint> endpoints )
+    bool GeniusNode::ConfigureRpcEndpoint( const std::string &chain_id, std::vector<WeightedRpcEndpoint> endpoints )
     {
-        if ( !transaction_manager_ )
+        auto transaction_manager = transaction_manager_;
```
EduMenges, line 2975: Why not just use the member?

EduMenges (approved):

PR #347, opened 2026-07-20
Coverage analysis

EduMenges:


src/account/GeniusNode.cpp
```diff
@@ -3524,7 +3525,12 @@ namespace sgns
         node_logger_->info( "Attempting reconnect to bootstrap fullnode {}...", peer_id.toBase58() );
 
         auto weak_self = weak_from_this();
-        pubsub_->GetHost()->connect(
+        auto ipv4_source = libp2p::multi::Multiaddress::create( "/ip4/0.0.0.0/tcp/0" ).value();
```
henriqueaklein, line 3528: Was this tested?

src/account/TransactionManager.cpp
```diff
@@ -1912,16 +1820,8 @@ namespace sgns
...
-            {
-                auto tx = GetTransactionByHashNoLock( input.txid_hash_.toReadableString() );
-                if ( tx )
-                {
-                    m_logger->debug( "Re-parsing {} transaction", tx->GetType() );
-                    BOOST_OUTCOME_TRY( ParseTransaction( tx ) );
-                }
-            }
-            account_m->GetUTXOManager().RollbackUTXOs( params.first, escrow_tx->GetHash() );
+            BOOST_OUTCOME_TRY(
```
henriqueaklein, line 1823: ParseTransaction was removed. Will this keep on working?

src/account/TransactionManager.cpp
```diff
@@ -1837,18 +1755,8 @@ namespace sgns
...
-            m_logger->debug( "Fetching transaction {} ", input.txid_hash_.toReadableString() );
-            auto tx = GetTransactionByHashNoLock( input.txid_hash_.toReadableString() );
-            if ( tx )
-            {
-                m_logger->debug( "Re-parsing {} transaction", tx->GetType() );
-                BOOST_OUTCOME_TRY( ParseTransaction( tx ) );
-            }
-        }
-        account_m->GetUTXOManager().RollbackUTXOs( transfer_tx->GetInputInfos(), transfer_tx->GetHash() );
+        BOOST_OUTCOME_TRY( account_m->GetUTXOManager().RestoreConsumedUTXOs( transfer_tx->GetInputInfos(),
```
henriqueaklein, line 1758: ParseTransaction was removed, will this work?

src/account/TransactionManager.cpp
```diff
@@ -2230,11 +2131,12 @@ namespace sgns
             {
                 ChangeState( State::READY );
             }
+            else
+            {
+                RequestRelevantHeads();
+            }
             return;
         }
```
henriqueaklein, line 2139: the "return" was removed. I don't think it should have been, as this request them actively and not through CRDT. Have you tested this?

henriqueaklein (changes requested):

src/account/GeniusNode.cpp
EduMenges: Yes, the idea is to use a different port when reconnecting due to port reuse bugs. We use the loopback address and the 0 port so the OS can pick one.

src/account/TransactionManager.cpp
EduMenges: But is it necessary to parse them if they're being rolled back?

src/account/TransactionManager.cpp
EduMenges: Hmm tbh this path is never reached, I'll add the return again as a safeguard

src/account/TransactionManager.cpp
henriqueaklein: Fair enough

src/account/TransactionManager.cpp
henriqueaklein: The idea is that when you have missing tx, you COULD fetch them individually, but now CRDT does it automatically and I don't remember if we fetch them while we are syncing we could end up with a broken CRDT state, that's why I kept it there

PR #348, opened 2026-07-21
Fix windows tests

EduMenges:


test/src/transaction_sync/genius_node_bootstrap_reconnect_test.cpp
```diff
@@ -57,6 +59,14 @@ namespace sgns
 
             bootstrap_address_ = full_node_->GetPubSub()->GetInterfaceAddress();
             ASSERT_FALSE( bootstrap_address_.empty() );
+            ASSERT_NE( full_node_->GetPubsubPort(), 0u );
```
henriqueaklein, line 62: there's a method for it, isn't there?

henriqueaklein (approved):

PR #349, opened 2026-07-23
Gsd/phase 09 canonical slot and certificate storage

henriqueaklein:


henriqueaklein:
This solution is bad and flaky. Not gonna merge it and instead gonna use as reference on what I want and how NOT to do it.

PR #350, opened 2026-07-23
Gsd/phase 09 securecrdt layer

henriqueaklein:


src/account/AccountMessenger.cpp
```diff
@@ -1026,7 +1026,8 @@ namespace sgns
 
     bool AccountMessenger::HasRequestPeers() const
     {
-        return pubsub_->getPeerCount( requests_topic_ ) != 0;
+        std::string topic = requests_topic_;
+        return pubsub_->getPeerCount( topic ) != 0;
```
EduMenges: Why?

src/multisig/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_library(multisig
+    MultiSig.cpp
+)
+target_link_libraries(multisig
+    PUBLIC
```
EduMenges:
```suggestion
    PRIVATE
```

src/multisig/MultiSig.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "account/GeniusAccount.hpp"
+
+namespace sgns::multisig
+{
+    bool VerifyPayloadSignature( const std::string          &address,
+                                 std::string_view            signature,
+                                 const std::vector<uint8_t> &payload )
+    {
+        return sgns::GeniusAccount::VerifySignature( address, signature, payload );
+    }
```
EduMenges, line 18: Is this method necessary?

src/multisig/MultiSig.cpp
```diff
@@ -0,0 +1,52 @@
...
+                continue; // invalid signature, skip silently
+            }
+            valid_unique_signers.insert( address );
+        }
+
+        QuorumResult result;
+        result.valid_unique_count = valid_unique_signers.size();
+        result.has_quorum         = valid_unique_signers.size() >= threshold;
+        return result;
+    }
```
EduMenges: Why is there a whole class just for this method?

src/securecrdt/ISignedCRDTData.hpp
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * @brief Interface implemented by every CRDT-backed value that requires a
+     *        quorum of authorized signers to create/update.
+     */
+    class ISignedCRDTData
+    {
+    public:
+        /**
+         * @brief Destroys the signed CRDT data instance.
+         */
```
EduMenges:
```suggestion
```

src/securecrdt/SecureCrdt.hpp
```diff
@@ -0,0 +1,159 @@
...
+         *        ISignedCRDTData subclass or Apply() side effect applies to a
+         *        given base_key -- that knowledge lives only with the
+         *        registered type's own owner (e.g. Phase 10 TrustedPeerRegistry,
+         *        Phase 11 BurnConfig, Phase 12 ValidatorRegistry migration).
+         * @param[in] base_key Registered CRDT key to read.
+         * @return outcome::success(bytes) if quorum is met, outcome::success(nullopt)
+         *         if the key does not exist yet or quorum is not yet met, or
+         *         Error::UNREGISTERED_KEY if base_key has no registry entry.
+         */
+        outcome::result<std::optional<sgns::base::Buffer>> ReadIfQuorum(
```
EduMenges, line 119: A result of an optional value?

src/securecrdt/SecureCrdt.cpp
```diff
@@ -0,0 +1,260 @@
...
+        const std::vector<uint8_t> payload = current_value.value().toVector();
+
+        auto sig_query = db_->QueryKeyValues( base_key.ChildString( "sig" ).GetKey() );
+        std::vector<std::pair<std::string, std::string>> collected_signatures;
+        if ( !sig_query.has_error() )
+        {
+            for ( const auto &[key_buffer, value_buffer] : sig_query.value() )
+            {
+                const std::string key_string( key_buffer.toString() );
+                const std::string address = LastKeySegment( key_string );
```
EduMenges, line 159: You are converting a key into a string, just for `LastKeySegment` to convert it back into a key again

src/securecrdt/SecureCrdt.cpp
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    std::optional<std::vector<sgns::crdt::pb::Element>> SecureCrdt::FilterSecureCrdtUpdate(
+        const sgns::crdt::HierarchicalKey &base_key, const SecureCrdtRegistryEntry &entry,
+        const sgns::crdt::pb::Element &element )
+    {
+        logger_->trace( "{}: entry key={}", __func__, element.key() );
+        const std::string address = LastKeySegment( element.key() );
+        const bool        is_signature_element = element.key() != base_key.GetKey() &&
+                                                  element.key() != ( "/" + base_key.GetKey() );
```
EduMenges: What the fuck

src/securecrdt/SecureCrdt.cpp
```diff
@@ -0,0 +1,260 @@
...
+
+        auto instance = entry.make_instance();
+        if ( !instance || !instance->DeserializeFromBytes( element_bytes ) || !instance->Verify( element_bytes ) )
+        {
+            logger_->error( "{}: malformed/invalid remote value rejected key={}", __func__, element.key() );
+            return std::vector<sgns::crdt::pb::Element>{};
+        }
+
+        logger_->debug( "{}: remote value accepted key={}", __func__, element.key() );
+        return std::nullopt;
```
EduMenges: What is the point in returning either a `std::nullopt` or an empty vector of elements?

src/securecrdt/SecureCrdtRegistry.hpp
```diff
@@ -0,0 +1,144 @@
...
+         *        regex shape (src/crdt/impl/crdt_data_filter.cpp).
+         * @param[in] key_pattern Base key pattern (regex-escaped by the caller
+         *            if it contains regex metacharacters).
+         * @param[in] entry Policy entry to register (compiled_pattern is
+         *            overwritten by this call).
+         */
+        static void Register( const std::string &key_pattern, SecureCrdtRegistryEntry entry )
+        {
+            entry.key_pattern     = key_pattern;
+            entry.compiled_pattern = std::regex( "/?" + key_pattern + "(/sig/.*)?" );
```
EduMenges: What is the point in this? Shouldn't the caller of `Register` already do this?

test/src/multiaccount/multi_account_sync.cpp
```diff
@@ -73,19 +74,20 @@ namespace sgns
...
-        static bool RemoveRegistryPersistence( const std::shared_ptr<GeniusNode> &node,
-                                               std::vector<uint8_t>               registry_block_key )
+        static std::string GetDatabasePath( const std::shared_ptr<GeniusNode> &node )
         {
-            if ( !node || !node->tx_globaldb_ )
-            {
-                return false;
-            }
+            return node ? node->write_base_path_ + node->gnus_network_full_path_ : std::string{};
+        }
```
EduMenges: This is only used once and likely `node` is always true, so it can be inlined.

test/src/multiaccount/multi_account_sync.cpp
```diff
@@ -73,19 +74,20 @@ namespace sgns
...
-            if ( !datastore )
+        static bool RemoveRegistryPersistence( const std::string &database_path,
+                                               std::vector<uint8_t> registry_block_key )
+        {
+            auto datastore_result = storage::rocksdb::create( database_path );
+            if ( datastore_result.has_error() )
             {
                 return false;
             }
+            auto datastore = std::move( datastore_result.value() );
```
EduMenges, line 90: What is the point in this change?

test/src/multiaccount/multi_account_sync.cpp
```diff
@@ -277,8 +279,13 @@ TEST_F( ValidatorRegistryTest, MissingRegistryBlockIsFetchedFromPeerByCid )
     ASSERT_TRUE( cid_bytes.has_value() );
 
     const auto client_base_path = GetBaseWritePath( node_client );
-    ASSERT_TRUE(
-        sgns::MultiAccountTestAccess::RemoveRegistryPersistence( node_client, std::move( cid_bytes.value() ) ) );
+    const auto client_database_path = sgns::MultiAccountTestAccess::GetDatabasePath( node_client );
+
+    client_registry.reset();
+    node_client.reset();
```
EduMenges, line 285: Is this for teardown fixes?

src/securecrdt/ISignedCRDTData.hpp
```diff
@@ -0,0 +1,65 @@
...
+        virtual bool Verify( const std::vector<uint8_t> &payload ) const = 0;
+
+        /**
+         * @brief Applies the side effect of this value once the caller has
+         *        independently confirmed quorum (e.g. via SecureCrdt::ReadIfQuorum).
+         *        Apply() itself does NOT check quorum.
+         */
+        virtual void Apply() = 0;
+    };
```
EduMenges, line 59: This class is only used in tests.

test/src/multisig/multisig_quorum_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+        "90bd26f57e3c243358666f32ff8321181545f4ddd8c981aceac163f26b05eaae",
+    };
+    // A sixth key, deliberately NOT in the signer set, used for the unauthorized-signer test.
+    constexpr char OUTSIDER_PRIVATE_KEY[] =
+        "90bd26f57e3c243358666f32ff8321181545f4ddd8c981aceac163f26b05eaaf";
+
+    std::string SignatureAsString( const std::vector<uint8_t> &signature )
+    {
+        return std::string( signature.begin(), signature.end() );
+    }
```
EduMenges: This function is called often, which makes me question whether `EvaluateQuorum` should accept a vector instead of a string in the first place.

test/src/multisig/multisig_quorum_test.cpp
```diff
@@ -0,0 +1,141 @@
...
+    std::vector<std::pair<std::string, std::string>> collected;
+    for ( size_t i = 0; i < 3; ++i )
+    {
+        collected.emplace_back( signers_[i]->GetAddress(), SignatureAsString( signers_[i]->Sign( payload_ ) ) );
+    }
+
+    auto result = multisig::EvaluateQuorum( signer_set_, 3, collected, payload_ );
+    EXPECT_TRUE( result.has_quorum );
+    EXPECT_EQ( result.valid_unique_count, 3u );
+}
```
EduMenges, line 74: WHy is the test named `ExactlyThresholdOfFiveHasQuorum` if you're using three?

test/src/multisig/multisig_verify_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+TEST_F( MultiSigVerifyTest, WrongSizeSignatureVerifiesFalse )
+{
+    auto account = GeniusAccount::NewFromPrivateKey( TokenID::FromBytes( { 0x00 } ), PRIVATE_KEY, path_ );
+    ASSERT_NE( account, nullptr );
+
+    const std::vector<uint8_t> payload = { 'p', 'a', 'y', 'l', 'o', 'a', 'd' };
+
+    EXPECT_FALSE( multisig::VerifyPayloadSignature( account->GetAddress(), "short", payload ) );
```
EduMenges: You're barely using the multisig here, since it is just a wrapper around the `GeniusAccount` signature operations. This whole test file could be deleted IMO.

test/src/securecrdt/securecrdt_interface_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+} // namespace
+
+TEST( SecureCrdtInterfaceTest, VerifyAcceptsMatchingPayload )
+{
+    TestSignedData data;
+    ASSERT_TRUE( data.DeserializeFromBytes( { 1, 2, 3 } ) );
+    EXPECT_TRUE( data.Verify( { 1, 2, 3 } ) );
+}
+
+TEST( SecureCrdtInterfaceTest, VerifyRejectsTamperedPayload )
```
EduMenges, line 52: What is the point of these tests?

test/src/securecrdt/securecrdt_quorum_contract_e2e_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        boost::filesystem::path                     path_;
+        std::vector<std::shared_ptr<GeniusAccount>> signers_;
+        std::vector<std::string>                    signer_set_;
+        int                                          token_ = 0;
+        std::unique_ptr<sgns::test::securecrdt::SecureCrdtTestNode> node_;
+        std::shared_ptr<SecureCrdt>                                secure_crdt_;
+    };
+} // namespace
```
EduMenges: Very similar to the fixture used in `securecrdt_propose_sign_quorum_test.cpp`, maybe they could be extracted into a common fixture

test/src/securecrdt/securecrdt_test_node.hpp
```diff
@@ -0,0 +1,156 @@
...
+                pubsub->Stop();
+            }
+            db.reset();
+            io.reset();
+        }
+    };
+
+    /// @brief Builds one unconnected GlobalDB test node named `dbName` for the
+    ///        currently-running gtest test case.
+    inline std::unique_ptr<SecureCrdtTestNode> MakeSecureCrdtTestNode( const std::string &dbName )
```
EduMenges, line 110: Wasn't there a fixture that already did this?

evmrelay
EduMenges, line 1: `checked_receipt_log_ordinal` is never used

EduMenges (changes requested):

src/account/AccountMessenger.cpp
henriqueaklein: Because it used to take a non const std::string reference and requests_topic_ was const.

src/multisig/MultiSig.cpp
henriqueaklein: well I didn't like that much either, but due to encapsulation it would make sense, as we would hide the definition and were we to change that, it wouldn't have any impact on callers. As this is a step on a multi phase design the idea would be to use this class from here on

src/multisig/MultiSig.cpp
henriqueaklein: The idea is to have a multi signature component that we can add more stuff if we need.

src/securecrdt/SecureCrdt.hpp
henriqueaklein: Well, yes. The idea here is that the result is to differentiate between an error (the key doesn't exist) from (a key exists, let's see if there's quorum in that).
The point of the SecureCRDT is to add a layer of lightweight consensus, so a key is inserted and then we wait for people to validate it by adding to quorum

src/securecrdt/SecureCrdt.cpp
henriqueaklein: It actually gets part of the key, which is the Last...Key...Segment

src/securecrdt/SecureCrdt.cpp
henriqueaklein: agreed

src/securecrdt/SecureCrdt.cpp
henriqueaklein: WHen it's a nullopt it means it's accepted, when not, it return extra key values that need to be removed as well. This was used so one element can remove other related elements. Scenarios:
1 - Correct data = nullopt, no data needs to be removed.
2 - Incorrect data = has_value(). Remove the element.
 2.a - If value is not empty, remove the elements that are returned as well.

src/securecrdt/SecureCrdtRegistry.hpp
henriqueaklein: No, the idea is that the user doesn't need to know the signature key pattern.
Let's say the caller wants to register that every key on "/tx/[^/]+" needs to be secureCRDT.
It's going to send  "/tx/[^/]+" as the key_pattern informing "that's the key scheme I want to be secure. Our SecureCRDTRegistry then inserts a  "/tx/[^/]+" + "(/sig/.*)?" as a signature entry pattern to filter for signatures.

test/src/multiaccount/multi_account_sync.cpp
henriqueaklein: yes.

src/securecrdt/ISignedCRDTData.hpp
henriqueaklein: yes, it's going to be used in future phases

test/src/multiaccount/multi_account_sync.cpp
henriqueaklein: The test would succeed sometimes for the wrong reasons, as it wouldn't fetch the registry from the node, it would grab from itself from the previous run.

test/src/multisig/multisig_quorum_test.cpp
henriqueaklein: it's N of M, so it's implicit that it's 3 out of 5. It's a shitty design by AI for sure. Gonna rewrite this BS

test/src/multisig/multisig_verify_test.cpp
henriqueaklein: it's an unfinished component. The initial design of it. The idea is to use multi sig and make it grow bigger

test/src/multisig/multisig_quorum_test.cpp
henriqueaklein: The core issue is that GeniusAccount has different types on sign and verify. Verify expects a string_view and sign returns a byte vector. I'm gonna add a verification method that expects bytes too, so we can rewrite this.

test/src/securecrdt/securecrdt_interface_test.cpp
henriqueaklein: We define a TestSignedData in which the verification of if being safe (signed) is just comparing the values. The idea is that a real derived class would have this using the multi sig

evmrelay
henriqueaklein: yes, it's used in another branch

test/src/securecrdt/securecrdt_quorum_contract_e2e_test.cpp
henriqueaklein: Agreed

test/src/securecrdt/securecrdt_test_node.hpp
henriqueaklein: did what? a SecureCRDT test node? I don't think so

src/securecrdt/SecureCrdt.cpp
```diff
@@ -218,43 +226,41 @@ namespace sgns::securecrdt
...
-        logger_->debug( "{}: remote value accepted key={}", __func__, element.key() );
+        const std::vector<uint8_t> payload = current_value.value().toVector();
+        if ( !multisig::VerifyPayloadSignature( address, element_bytes, payload ) )
+        {
+            logger_->error( "{}: invalid remote signature rejected key={}", __func__, element.key() );
+            return std::vector<sgns::crdt::pb::Element>{};
+        }
+        logger_->debug( "{}: remote signature accepted key={}", __func__, element.key() );
         return std::nullopt;
     }
```
EduMenges, line 265: It still is not clear to me as to why you return an empty vector

EduMenges (approved):

src/securecrdt/SecureCrdt.cpp
henriqueaklein: the idea would be to return the elements that would need to be erased. So we could return the input always and add any extra elements we want to delete, that would be easier to understand but require some extra work on the caller, which will have to see if the first element matches the input. Seems unecessary, but it might be the best way to, although it falls out of scope here

